PR #1, opened 2019-04-24
Single page application

internettrans:


text/0001-single-page-applications.md
```diff
@@ -0,0 +1,34 @@
+# Single Page Applications
+- Start Date: 04/23/2019
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
+
+## Decision, including impact on distributions
+Core OpenMRS frontend modules will be implemented within a master single page application. Distributions are encouraged
```
dkayiwa, line 6: I guess this takes care of OpenMRS Web Apps. Correct?

https://wiki.openmrs.org/x/C4KIBQ

https://wiki.openmrs.org/x/_o2QBQ

internettrans: Oo thanks for linking me to those -- I hadn't seen them and didn't know about packaged web apps in OpenMRS. Is this how distributions like Bahmni and Ampath's distro are writing single page applications right now? Creating their own packaged web app?

It looks like packaged web apps have their own HTML file -- my proposal in this PR is to consolidate all HTML files into one master single page application. Modules that plug into that master single page application can have javascript, css, img, and font assets. But they would not be able to have their own HTML file.

Thoughts on that? ^

dkayiwa: That looks fine to me. Bahmni and Ampath are not Open Web Apps. I just mentioned it for a few Open Web Apps that are already in existence. Though not as many as the JSP, GSP ones.

text/0001-single-page-applications.md
```diff
@@ -0,0 +1,34 @@
...
+to add modules to the master single page application. Older OpenMRS modules implemented with JSPs and GSPs are not addressed in this
+proposal, but will be covered in another one.
+
+## Definition
+A [single page application (SPA)](https://en.wikipedia.org/wiki/Single-page_application) is a frontend application that does not
```
dkayiwa, line 13: We also had what we call user interface extension points where modules would extend the default user interface at various points as needed by means of server side java files. An example would be extending the patient dashboard with more tabs. Do you think this will blend very well with SPA? Am almost sure it will. Just a sanity check. 😊  https://wiki.openmrs.org/x/sA8z

internettrans: Hmm, not sure about this one. One of the biggest differences between server rendered JSPs/GSPs and client rendered SPAs is that the modularity/extensibility is happening *in the browser* instead of in the Java code.

Reading through the Module Extension Points wiki that you linked to, those are very much tied to JSPs and are run on the server, not in the browser. Each JSP is its own HTML file, but in an SPA architecture there is only one HTML file and it's not doing any complex rendering on the server side because it lets the browser-side javascript do all the work.

So I think that user interface extension points (as explained in the wiki) would not be part of the new architecture, but rather something that would only continue to be supported for older code. This is related to the **big** question of what to do with the JSPs and GSPs in terms of a migration path. I intentionally left that out of this RFC PR in favor of trying to tackle that one in a subsequent PR.

dkayiwa: Do you plan having an equivalent of new extension points from the browser side?

internettrans: Let me make sure I am understanding Module Extension Points -- they allow you to modify the HTML that gets sent down for a specific page?

There will not be an exact equivalent to these. The proposed architecture is one where there are core OpenMRS microfrontends that distributions may choose to use, and then additional microfrontends that are developed by individual distributions. The ability to pick and choose which microfrontends you want, and to add in your own microfrontends, is what I have been thinking as the main way to customize/extend the OpenMRS frontend. It sounds like Module Extension Points is a way to **modify** core OpenMRS frontend JSP pages, not just pick and choose which ones to use. In the new architecture, I do not have a proposal in mind for letting people modify a microfrontend's code on-the-fly -- I was only thinking in terms of allowing them to pick and choose which microfrontends to include or not, and then add their own microfrontends.

Does that make sense?

dkayiwa: Yes it makes sense.
Is this just hard to accomplish with the proposed architecture?

text/0001-single-page-applications.md
mogoodrich: Is the idea that the *all* distributions will include a giant SPA that includes *all* the the front ends built by all contributing distributions, or will there be a "pick-and-choose" model where each distribution will build and compile their own SPA based on the modules they want to use?

mogoodrich: Update... sounds like reading further comment the answer is the latter, the "pick-and-choose" model...we should probably make that clear in this doc.

text/0001-single-page-applications.md
mogoodrich: An example of how we currently use extension points:

On my distro's clinician dashboard (or any dashboard) say I want to display a set of patient summary widgets in some sort of order.. to do this currently one can use some json config to "inject" those widgets into a specific "patient dashboard" extension point.  So in the new model, suppose you have an "current lab test" widget, a "current medications" widget, and a "current vitals" widget, and they are provided by different microfrontends, would there be a mechanism for including them on a single "dashboard" page?

A simpler extension point example is "Visit Actions" and "Actions" extension which is basically a list of actions that a user can take in the current context, and generally each action maps to a link... continuing from the example above you might have have "Order Tests", "Order Labs" and "Enter Vitals" actions that take the user to specific microfrontends for ordering test, labs, entering vitals, etc... ie we are not trying to combine the display from multiple front ends on a single page, but rather allowing the user to navigate to UIs provided by different microfrontends within a certain patient context.

text/0001-single-page-applications.md
internettrans: Yeah, the pick and choose model. Some of the technical details are not fully fleshed out, but a lot of it is. We are proposing to use [modules](https://github.com/openmrs/openmrs-rfc-frontend/pull/2), [import maps](https://github.com/WICG/import-maps), and [single spa applications](https://single-spa.js.org/docs/building-applications.html) to structure the code. Distributions will be able to author their own single spa applications that are modules that get put into the import map.

Some core modules will be baked into the master single page application, but after that it will be up to the distributions to add their own modules.

jonathandick: @joeldenning I suspect it is unclear to some (at least me) one it means to be a "single-spa application" as well as the alternative options to not using single-spa.  It may help to provide more commentary on this (perhaps it belongs in it's own PR which recommends single-spa to be the router of routers?) Or not... I'll defer to you.

text/0001-single-page-applications.md
internettrans: @dkayiwa 

> Is this just hard to accomplish with the proposed architecture?

Module extension points seem to be "throw in some javascript and html into a jsp," which means that you can do almost anything with them but also that there's no structure to how they interact and it is harder to ensure that they work together when there are multiple extensions within a single jsp.

With the new architecture, you still can "throw in some javascript and html into the master single page application," but we will be encouraging people to follow the structure and architecture in this RFC instead of reaching into other modules dom and doing stuff with it. There will be ways to configure and modify the behavior of core pages and code.

@mogoodrich thanks for the examples of extension points.

The answer to how to accomplish that kind of thing in the new architecture is different depending on whether it is modifying an existing page that is in openmrs core versus adding a new page.

**Modifying openmrs-core functionality**: For this, I think there's some exploration to do. I'm open to ideas/proposals. One option would be to pass in [single spa custom props](https://single-spa.js.org/docs/building-applications.html#custom-props) to the core modules that change their behavior, or even add your components to the module.

**Adding brand new functionality**: For this, distributions will be able to author their own [single-spa applications](https://single-spa.js.org/docs/building-applications.html) that can take over specific routes. Then the distribution's application code can fully control that page, perhaps reusing some prebuilt components exposed by OpenMRS.

Does that better explain what this proposed architecture will offer?

text/0001-single-page-applications.md
```diff
@@ -0,0 +1,34 @@
...
+to retrieve data from the backend.
+
+## Reason for decision
+- SPAs are the industry standard for large web apps.
+- SPAs are often faster for users that visit many pages within an app. You pay an upfront performance cost with the initial load time, in order to avoid
+subsequent page reloads.
+- The tooling and developer experience for SPAs is preferred over server rendered pages.
+
+## Alternatives
+The alternative to SPAs is server rendered HTML files for each frontend route. This means that the page reloads whenever the user navigates.
```
djazayeri: There's a key alternative missing here, which is to have SPA-per-"app", rather than one single SPA.

The home screen would be a switchboard of apps, each of which is fundamentally just a URL. An app can be implemented in any tech stack, and they don't have to share any code or state. (This provides a way to expose legacy server-side functionality).

Going forward we'd hope that most new code gets written in modern technologies with modern style, and maybe people can start to collaborate around shared JS code by introducing an "openmrs-shared.js" library that big dev players commit to start sharing; it would starts out empty, and gets built up as people collectively agree on what to put in there. In this model we don't necessarily need single-spa, as individual apps would probably still be built in either React or Angular, and don't need to mix them.

Though there's also a version of this model where we have a single-spa based application that includes all stuff built going forwards. This would make sense if there are actually large-scale contributors who want to share things, but can't agree on React vs Angular. (It would also be compelling if this were a way to bring the Bahmni team into the fold, which is mostly AngularJS.)

The key point is to enable a slower migration story. I speculate that many of the organizations who have built their own distributions are more able to adopt new things one clinical or programmatic area at a time, or maybe they want to experiment with a shiny new tool in a new program with just a few users, without breaking their existing EMR setup or requiring retraining. Maybe they are locked into some components that they've customized heavily, but have more flexibility to replace others. For example an implementation might have customized the (GSP-era) registration app, and has no interest in moving from this. But they're excited about a new Clinical app, and they'd invest in adopting and developing this one.

Big picture, the hardest things to solve are (1) existing large amounts of GSP and AngularJS 1.x code, (2) path for  incremental adoption of any new thing, assuming that it's going to take quite a long time to fully build given the available dev capacity. So I would focus on these as the top priorities; as it is they feel a bit incidental to this design.

mogoodrich: Great summary @djazayeri , and reflects PIH's current situation well.

djazayeri: Having slept on it, the other key reality in the OpenMRS world is that there's never enough available dev time to "rewrite the whole thing in modern framework X", so this _is_ an argument for imposing requirements up-front to how individual pieces of work are structured, so that JS frameworks can coexist. That's a strong argument for creating new conventions that let _new_ code be assembled without specifying a framework.

I understand that one of the drivers here is that AMPATH prefers Angular, while PIH prefers React. I'm curious to hear what we're hoping to actually achieve with React and Angular code running side-by-side. I can think of several possible things, of varying difficulty (and some have bad side-effects):

1. someone can assemble the AMPATH HIV app with the PIH TB app (but each app uses just one framework)
* AMPATH and PIH can both contribute to a new NCD app that will mix Angular and React code
* have a shared patient dashboard composed of widgets; these widgets can use any modern JS framework
* share browser-side data across apps/components of different tech stacks
* don't want to have to implement identical global nav in both Angular and React

jonathandick: @djazayeri  thanks for these thoughtful comments. I will defer to Joel and others to review the pros and cons of various migration strategies from legacy systems. 

I feel that we need to solve two pressing concerns. 

One, we lack a convention for how to collaborate in the client side javascript world. I think this is the most common way of building applications these days (i don't think jsp/gsp's are used too frequently) and we ought to figure out how to solve this problem. Microfrontends seems like a viable approach and hence this discussion. 

Two, we need a way to upgrade parts of the application without needing to upgrade everything. For example, moving from angular 4 to angular 6 took us about 6 months because we needed to upgrade everything. If  we could somehow break thinks into smaller widgets that could be upgraded as needed (e..g a vitals widget in angular 6 and a lab results widget in angular 7), this would make it much easier to keep up to date make use of the latest advances in various frameworks or perhaps switch to a new one. 

AMPATH would benefit from using a microfrontend approach independent of whether other groups collaborated with us. This is by no means saying we are looking to go it alone in developing an application framework (in fact, in two paragraphs, I argue that we can only survive if we find away to work collaboratively) . Only, that we have already experienced the pain of having a large monolith and would like to avoid having to go through that again. 

The other reality is that we (ampath) are building one widget after another which is by no means ampath specific but more core features of an EMR. For example, an order entry module, an appointment scheduler, lab results, vitals, and the list goes on. I believe others are doing the same. And as demand for our EMR grows, we are having trouble both maintaining and developing new features. I believe others are also having this problem. **Finding a way to collaborate which addresses both concerns above is our highest priority.** 

Lastly, as we are just beginning this process, it's not clear how much functionality will exist within a microfrontend. If at the end of the day, the microfrontend has large amounts of functionality, e.g. an entire patient dasbhoard i(ncluding widgets to view encounters, visits, vitals, labs, create notes, create orders, etc.) then I think collaboration will be tricky. However, if a microfrontend consists of only a vitals widget or only a lab results widget (and we have a style guide), then it's much easier to imagine having an app store/module repository where you pull in the widgets you want and replace them as needed. This is where I'm hoping we're headed. And i think there is a migration pathway there where as new widgets become available that your implementation wants, you include them in your build. If you prefer to continue to use a legacy approach that can not fit with this single-spa framework, then you simply have links to move back and forth.

internettrans: @djazayeri thanks for bringing up these good points -- you are totally correct that doing multiple SPAs is an alternative that is not explored or discussed here. And that it should be.

I've done multiple SPAs before at other companies, and it's a viable solution. I think it ultimately comes down to which set of pros and cons we want. Here are the pros and cons of one SPA vs multiple SPAs, as I understand it:

## One SPA
#### Pros
- You do not have to refresh the page when navigating between features. This is huge when there are two interrelated features where it's a pain to have to always reload when going back and forth. The app is faster when you don't have to.
- You can have more than one microfrontend active at a time.
- It is easier to share logic, components, and styles within a single page vs across multiple pages. You don't have to worry about having different versions of the navbar or styleguide or other shared components/utils whenever you navigate to a new feature.
- Moving functionality from one codebase to another codebase is easier, since the code still is in the same runtime and has the same available global variables, in-browser modules, fonts, styles, etc.
- Links and navigation are always using pushState, instead of sometimes being pushState and sometimes being a full page reload. It's a bit of cognitive overhead for each developer to know whether to cause a page reload or not, so it's nice just to not ever worry about knowing when you're doing one or the other.
- It is easier to configure the server -- all frontend routes hit a single html file, instead of routing having to happen both on client side and on server side (to determine which html file to render)

#### Cons
- In order to get microfrontends to share an html page, you often have to modify their build process such that they no longer generate their own html page. This initial, upfront work can be a pain
- You have to introduce a new thing, a top-level router, that controls which features/microfrontends are active. When doing multiple SPAs, you're letting the page reload act as the "top level router," which is an understood and reliable technology.
- The boundaries between different features/microfrontends can become blurred if you're not careful. If left unchecked, this can result in tight coupling and low cohesion for each of the microfrontends.
- Not ever reloading the page exposes existing memory leaks in the code more than occasionally reloading the page does. This is because a page reload causes a garbage collect that can hide memory problems.
- You have more javascript code running within a single browser tab, which can result in that tab using more memory.

## Multiple SPAs
#### Pros
- Every microfrontend can completely control its own html file.
- The microfrontends are more isolated -- they do not share a javascript runtime, DOM, etc.
- Multiple SPAs is probably a more-understood architecture than sharing a single SPA for multiple microfrontends
- (see some of the cons of one SPA for some more pros of multiple SPAs)

#### Cons
- You can only have one microfrontend active at a time
- Sharing logic, data, components, and styles is more difficult. You have to version all of them and then try to get all of the separate html files to upgrade independently.
- Much harder to get new code pushed out to and used by all the microfrontends
- More complex server routing. Both the server and the client have to know about the top level routes.
- (see some of the pros of one SPA).

internettrans: Maybe let's discuss this further over chat? I'm in the OpenMRS slack channel and am happy to talk about it with you -- a 100% open conversation where I'd be open to learning and discussing and changing the direction that we're going here.

internettrans: > I understand that one of the drivers here is that AMPATH prefers Angular, while PIH prefers React.

^ I actually don't view this as the reason. I think the bigger reason is that we want to have an architecture that doesn't need to be redone every couple years, and that betting on any single framework in the frontend world is tricky -- AngularJS was king 4 years ago, but is now dead. I'm hoping to avoid that situation happening to the new OpenMRS frontend.

> The key point is to enable a slower migration story. I speculate that many of the organizations who have built their own distributions are more able to adopt new things one clinical or programmatic area at a time, or maybe they want to experiment with a shiny new tool in a new program with just a few users, without breaking their existing EMR setup or requiring retraining. Maybe they are locked into some components that they've customized heavily, but have more flexibility to replace others. For example an implementation might have customized the (GSP-era) registration app, and has no interest in moving from this. But they're excited about a new Clinical app, and they'd invest in adopting and developing this one.

Could you help me understand how multiple SPAs would allow for a slower/more incremental migration strategy? My plan with single-spa is to make Bahmni a single-spa application *as is*, without having to do a big migration. Same goes for Ampath, PIH, etc. Those who are already doing client rendering can keep their code and just fit it into the SPA.

My fear with the suggestion you made of a home page with some links to Bahmni, Ampath, PIH, etc, is that there would be large overlap between those applications' feature sets and that it wouldn't feel like a unified EMR app to link to each of them individually. Having a link to each might just feel like a conglomeration of competing UIs. I might be misunderstanding the intent of that home page though.

text/0001-single-page-applications.md
internettrans: @jonathandick my plan is to have a separate RFC proposal that goes over single-spa applications. You are right that it's a top-level router -- another way of thinking about it is that single-spa will provide structure to the modules in our SPA.

text/0001-single-page-applications.md
mogoodrich: What I'm having a hard time seeing is how PIH EMR, or the Ref App, which has a huge smorgasbord of server-side code, could "just fit into the SPA" without having do a big migration.  If that does turn out to be easy, then, yes, there is less value in multiple SPA approach.  So, admittedy, I'm skeptical, but happy to be proven wrong... :)  For existing server-side based implementations, would the idea be that the single SPA just start out operating as a lightweight router device, with all the existing page still being served server-side underneath?

internettrans: > What I'm having a hard time seeing is how PIH EMR, or the Ref App, which has a huge smorgasbord of server-side code, could "just fit into the SPA" without having do a big migration

Yeah I do too :). On [line 8](https://github.com/openmrs/openmrs-rfc-frontend/pull/1/files#diff-ea28b96317944fb0854c7d6d23c7b47aR8) I wrote "Older OpenMRS modules implemented with JSPs and GSPs are not addressed in this proposal, but will be covered in another one," which essentially punts on the whole topic for now, with a future RFC PR talking about the plan for JSPs. I have some ideas in mind for how we could migrate from server rendered to client rendered, but that will definitely be much harder to do in small increments. UIs that are already client rendered, such as Bahmni and Ampath, will have a much easier time migrating.

Switching from server rendered to client rendered is the single biggest change that I'm proposing, so I'm glad you're calling it out. The bet is on client rendered SPA as the best choice for OpenMRS' future. And that by making the switch to client rendering, we'll be able to use modern frameworks and tooling while hopefully avoiding a situation where we're cornered into a rewrite again. But the switch from server rendered to client rendered will definitely involve a fair amount of rewriting -- we'll have to develop a strategy for legacyui and reference implementation

mogoodrich: Cool, thanks @joeldenning that makes sense and I wouldn't expect there to be an easy migration path... but PIH makes heavy use in the old server-side technologies (and so pretty much all implementations beyond AMPATH and Bahmni-based ones), so if the idea is that we wouldn't be able to take advantage of this new SPA until we migrate all our old code (not sure if that it the case) that would be *way* down the road for us and we'd have a hard time investing our limited resources in the effort at this point... what @djazayeri is suggesting provides a clearer path for how I could see us utilizing something that comes out of this initiative in the short-to-mid-term...

internettrans: One of the options I see for migration from server rendered pages to SPA is to keep the server rendered pages working, but have the landing page for the UI be the SPA. The SPA will start out without feature parity and would do lots of linking to ref app / legacyui / other jsps and do a page reload to navigate to them. Then, over time, we can rewrite and replace any parts of ref app / legacyui / jsps that we want to move into the SPA.

Thoughts on that? ^

I think that what I described might be similar to what @djazayeri is proposing, except that there would only be one SPA + many server rendered pages, instead of many SPAs. I don't think any proposal could make sense that immediately and completely throws away the JSPs/GSPs -- there will definitely have to be compatibility with the JSPs and a path for migrating away from them slowly (or maybe never). But for newer code and distributions, the hope is that they would be able to take advantage of the benefits of modern frameworks and SPAs. With a clear path forward toward a cohesive, extensible user experience that many or all distributions can use.

mogoodrich: Yes, that makes sense and is definitely worth exploring!

dkayiwa: Despite the recommended client side rendering, there could still be developers out there who still prefer server side rendering, even for new functionality or applications. Therefore, i would not look at this just as a mere upgrade path, but rather support for server side rendering in the new architecture.

djazayeri: > Maybe let's discuss this further over chat? I'm in the OpenMRS slack channel and am happy to talk about it with you -- a 100% open conversation where I'd be open to learning and discussing and changing the direction that we're going here.

Is there an OpenMRS slack channel that I don't know about? Or do you mean IRC/Telegram? Anyway, we can schedule a time to chat someday if need be.

For now I think maybe we're actually saying the same thing in different words. Let me try a different emphasis:

* If major dev players can agree to collaborate on future development by building one SPA composed of microfrontends, that's 100% good. (I'll call this the main SPA.)
* It's worth sacrificing (a lot of) dev velocity today if we're 100% sure this will make it easier to incrementally upgrade frameworks or swap out tech stacks in the future.
* Some form of backwards-compatibility is blocker-level important; but this doesn't need to be pretty. (Okay to do full page loads. Okay if navigating from the new SPA to legacy GSP has some UI inconsistencies.)
* I recommend handling backwards compatibility by still having a home screen that looks like "apps" and some of those apps are just external links out of the main SPA. (At first there may be many of these, eventually few or none.)
* The main SPA should also support the idea that even going forward, devs might want to write functionality outside of it, for reasons that may be good or bad, in technologies that may be client-side or server-side.
    * It's okay for this "support" to just means "we provide a mechanism for having 'external' links (from the main SPA) to what the end-user would consider part of the overall application. And we are prepared for 'incoming' links, e.g. someone might write a custom patient search in another tech, but once the user picks a patient it links back to the main SPA's patient dashboard.
    * I'm not really suggesting anything different technically, this is more about expectations and how-to documentation.
    * One thing implied here (but not specific to this) is that the main SPA should have good treatment of "admin-configured lists of links", which can potentially point to arbitrary URLs. E.g. the admin configures what "shortcut actions" are available from the patient dashboard, and each of these things is a title+(internal route | external url). This covers one of the main use cases of "extension points" that Daniel was mentioning before.

Trying to summarize that further:

_Take the "one SPA of microfrontends" approach for future code _but_ assume people want to bolt on other code that's part of the same end-user-facing app by cross-linking between their code and the main SPA._

djazayeri: A separate point (not sure where to address this)...
I got the quick impression that one of the benefits of single-spa specifically is to be able to compose JS modules from different tech stacks in arbitrary ways. I'm not sure that completely compositions is so valuable.

I would speculate that it's okay to make a decision like "all the global nav and routing will be written in (react/angular)". The highest value is if individual routes can be js modules of any modern js library, and also if certain pages like a patient dashboard can use components from any modern js library.

internettrans: > Is there an OpenMRS slack channel that I don't know about? Or do you mean IRC/Telegram? Anyway, we can schedule a time to chat someday if need be.

Yeah there is an OpenMRS slack workspace. It's not meant to be secret in any way -- I just sent you an invite email at jazayeri@alum.mit.edu. Let me know if you'd prefer a different email. Also, I'm not yet on the IRC but we could also talk there.

> For now I think maybe we're actually saying the same thing in different words. Let me try a different emphasis:

Yeah I think so, too. One area I haven't focused on as much is the incremental and slow migration path for the large amount of server rendered code, which is something I do think is very important and shouldn't be dismissed. That might have been the root of any alarm or disagreement.

> If major dev players can agree to collaborate on future development by building one SPA composed of microfrontends, that's 100% good. (I'll call this the main SPA.)

Agreed. There is a meeting on Monday morning EST that Burke is leading where this will be discussed. Definitely feel free to join.

> It's worth sacrificing (a lot of) dev velocity today if we're 100% sure this will make it easier to incrementally upgrade frameworks or swap out tech stacks in the future.

💯 

> Some form of backwards-compatibility is blocker-level important; but this doesn't need to be pretty. (Okay to do full page loads. Okay if navigating from the new SPA to legacy GSP has some UI inconsistencies.)

Agreed. I will be submitting an early, rough draft RFC PR in the next day or two to try to tackle this, but would appreciate your feedback and collaboration on it, since you know much more about the reference implementation and server code than I do.

> I recommend handling backwards compatibility by still having a home screen that looks like "apps" and some of those apps are just external links out of the main SPA. (At first there may be many of these, eventually few or none.)

Good proposal, let's discuss this in the rough draft RFC PR that I'll submit (mentioned above).

> The main SPA should also support the idea that even going forward, devs might want to write functionality outside of it, for reasons that may be good or bad, in technologies that may be client-side or server-side. 

I don't think there are plans to ever entirely remove the ability to write server rendered html pages entirely. My vision for it is that all documentation and easy paths will point toward SPA, but that JSP/GSP will be possible if you want to go rogue / stay with older technologies.

> I got the quick impression that one of the benefits of single-spa specifically is to be able to compose JS modules from different tech stacks in arbitrary ways. I'm not sure that completely compositions is so valuable.

That is one of the primary benefits, although there are other benefits, too, such as allowing for separate modules to have their own build processes and to be more decoupled than if they shared a build process. I personally believe that remaining flexible to more than just one framework is crucial to avoiding large rewrites in the future -- happy to elaborate on this or discuss further if you'd like.

> I would speculate that it's okay to make a decision like "all the global nav and routing will be written in (react/angular)". The highest value is if individual routes can be js modules of any modern js library, and also if certain pages like a patient dashboard can use components from any modern js library.

I agree that global nav would be an app that has only one framework, but disagree that all routing would have to be only one framework. In other words, the nav would use a framework to implement its components, but it wouldn't require the main content apps to use the same router or framework when implementing their components. If all routing is one framework, that means that that framework is king, since the router is ultimately in charge of everything that gets rendered.

djazayeri: > I agree that global nav would be an app that has only one framework, but disagree that all routing would have to be only one framework.

You're correct. I phrased this badly, but I think we agree.

internettrans: I just pushed a commit to this PR that adds multiple SPAs as a listed alternative. I also linked to this github comment thread in the markdown for anyone interested in reading this later on.

We are closing in on merging this PR -- see https://talk.openmrs.org/t/micro-frontends-architecture-for-openmrs/22557/9 for notes from the first "microfrontends squad" meeting.

angshu:
Excuse me for not replying inline, writing from mobile browser is not fun. 
Regarding single-spa, I have some reservations.
- typically a type of persona will use only one type of app. E.g registration, clinical, ipd, appointments. Why load all the other apps? I would prefer that we have multiple spas - reduces load time, complexity and has better manageability, imho. And makes app/control sharing contextual. 
- less code, more cohesion that way. 
- the normal tendency in openmrs has been, write everything in openmrs (e.g lab, radiology). I believe that if we allow for multiple spas, it will make people segregate things 

Another aspect is owa!
- it's a dead spec. Let's not use it as excuse for packaging
- people should be free to distribute as they like. Pack in owa, or distribute through httpd, nginx etc. 
- taking it outside of tomcat also forces people to think of better way of managing code.
- typically you would think of 2 layers. Front end totally separate and deployable than server APIs. I should be able to deploy my API, without touching or redeploying front end code
- imagine a future state, where backend is also microservices. Why deploy frontend on a Omod! 
- imagine backend API/apps in cluster, multiple nodes. I wouldn't want to redeploy my front end in all nodes.

internettrans:
@angshu thanks for your comments:

> Regarding single-spa, I have some reservations. Typically a type of persona will use only one type of app. E.g registration, clinical, ipd, appointments. Why load all the other apps?

I think you're fundamentally misunderstanding single-spa. You do not load all the apps. You only load the apps that you navigate to. The vast majority of code will not be loaded until you navigate to the route that needs it. With single-spa, applications that are not used are not downloaded. See https://single-spa.js.org/docs/configuration.html#loading-function

>  I would prefer that we have multiple spas - reduces load time, complexity and has better manageability, imho. And makes app/control sharing contextual.

I agree that multiple spas is a valid alternative. See https://github.com/openmrs/openmrs-rfc-frontend/pull/1#discussion_r280165194 for my analysis of Pros and Cons.

I disagree that multiple spas reduces load time -- in fact I believe the opposite. Multiple SPAs means you're constantly reloading any shared code as you navigate between the apps. Lots of SPA page reloads is a bad user experience, since each SPA is slower on initial load than a server rendered page.

I agree that multiple SPAs reduces complexity in some ways, but think it introduces complexity in other ways. One example of increased complexity for multiple SPAs is that dependency management becomes a lot harder -- how do you ensure that every SPA has the same version of the auth library or navbar? Another example of increased complexity is that every link has to know whether it is a full page reload or not -- should it call `history.pushState()` vs `location.href = '/new-url'`. This is something that every dev has to think about whenever they make a link.

> And makes app/control sharing contextual.

I don't understand what you mean by this, could you clarify?

> the normal tendency in openmrs has been, write everything in openmrs (e.g lab, radiology). I believe that if we allow for multiple spas, it will make people segregate things

I think you're misunderstanding the proposed architecture, and the role openmrs-module-spa. We are not going to have one big frontend codebase. Instead, there will be lots of small repos that have their own build processes. They will not be java modules, and each will have their own package.json and webpack config.

> Another aspect is owa! it's a dead spec. Let's not use it as excuse for packaging.

I 100% agree. New architecture does not use OWAs in any way.

> people should be free to distribute as they like. Pack in owa, or distribute through httpd, nginx etc.

Agreed, and planned for

> taking it outside of tomcat also forces people to think of better way of managing code.

My preference is for Tomcat to control the root html file, but not the js files. Reason for this is that it guarantees that all openmrs users can actually use the root html file. For the js files, my current thinking is to allow distribution maintainers to choose whether to use Tomcat or not, with a preference towards distributions using a CDN or nginx.

> taking it outside of tomcat also forces people to think of better way of managing code.

I agree that Tomcat causes code to be crazy. With my proposed architecture **you won't even have to boot up openmrs-core in order to develop frontend code**.  In regards to whether deployed javascript code is hosted via tomcat or not, see my previous comment where I think it should be left up to the implementor and not a decision made for all OpenMRS users.

> typically you would think of 2 layers. Front end totally separate and deployable than server APIs. I should be able to deploy my API, without touching or redeploying front end code

Yep -- that is 100% what I'm going for. Deploying frontend code will not require an openmrs-core or openmrs-module-spa deployment. Instead, deployment of frontend code will involve uploading a js file to a CDN and then updating the [import map](https://github.com/openmrs/openmrs-rfc-frontend/pull/4) to point to the new file. 

> imagine a future state, where backend is also microservices. Why deploy frontend on a Omod!

Agreed -- I'm not sure if you're saying that my proposed architecture would require that, but to clarify a backend deployment is 100% separate from frontend deployments in my architecture.

> imagine backend API/apps in cluster, multiple nodes. I wouldn't want to redeploy my front end in all nodes.

Agreed -- I think you've misunderstood the proposal because that is absolutely what I'm going for.

@angshu I appreciate your thoughts -- I honestly believe that many of your comments are based in a misunderstanding of the proposal and think we're pretty aligned on what an ideal frontend architecture looks like. Feel free to message me on OpenMRS slack or respond to this github comment with further questions.

internettrans:
To add a bit more context to the multiple SPA vs one SPA question — I have implemented multiple SPAs before and then migrated off of it because it is a poor user and developer experience. And now I’m currently helping three companies migrate away from multiple SPAs for the same reasons — they are paying me to combine the SPAs via single-spa because the multiple SPA architecture didn’t work out.

So my preference is to not have OpenMRS go down the same way path 😀

fatmali:
To add to the single-spa vs multiple-spa discussion, I can see the need for having multiple-spa's for the different apps that currently exist in legacy. But do we really need them to be separate multiple-spas? From what I understand, single spa's don't have to share any data and only one can be visible given a certain route. Each app in legacy can then be a separate route within the a single-spa. Multiple-spa sounds like an anti-pattern, and would lead to future pains when it comes to maintenance (a problem that we're trying to solve using single-spa).

internettrans:
@fatmali @angshu @ojwanganto Could you take a look at this and either comment or approve? To approve, you can add your Thumbs Up emoji, make a github comment saying you approve, or doing a real Github approval review if you have permissions to do so.

I'm waiting on 3 core team members approving this before merging. We already have one from @brandones

fatmali:
@joeldenning  I approve.

internettrans:
Three approvals, now, so I'll merge

PR #2, opened 2019-04-25
Proposal for javascript modules

internettrans:

### DIFF
--- a/text/0002-modules.md
+++ b/text/0002-modules.md
@@ -1,0 +1,57 @@
+# Modules
+- Start Date: 2019/04/29
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/2
+
+## Decision, including impact on distributions
+In-browser Javascript modules are the OpenMRS mechanism for adding functionality and creating frontend code for distributions.
+Since modules are not yet fully supported in all browsers, OpenMRS uses SystemJS as its in-browser module loader.
+
+## Definition
+A [javascript module](https://www.sitepoint.com/understanding-es6-modules/) is a new kind of javascript file that
+can [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and
+[export](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) data and logic from other modules.
+Variables within a module are not global-by-default and are not accessible to other javascript code unless explicitly
+exported.
+
+Javascript modules are also known as ECMAScript modules, ES modules, ESM, ES2015 modules, or ES6 modules.
+
+An in-browser javascript module is a javascript module that is loaded in the browser. This is different from
+build-time modules, which are compiled down so that they know longer use `import` or `export` by the time they
+are executed in the browser. Most major frameworks (React, Vue, Angular, etc) always use build-time modules, but do not
+always use in-browser modules. At OpenMRS, we use both.
+
+In-browser javascript modules are downloaded with one network request per module. Build-time modules are compiled away such
+that there is only one network request for a whole group (bundle) of build-time modules.
+
+[Browser support for modules](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Browser_compatibility)
+started in Chrome 61, Firefox 60, Edge 16, and Safari 10.1. Internet Explorer does not support modules.
+
+A module loader is a javascript library that allows you to use modules in a browser that does not support them. This makes
+it possible for us to use modules in Internet Explorer.
+
+[SystemJS](https://github.com/systemjs/systemjs) is a minimalistic module loader that
+[polyfills](https://en.wikipedia.org/wiki/Polyfill_(programming)) modules so that you can use them in older browsers.
+When browsers fully support modules, the browser will be able to do what SystemJS is currently doing for us.
+
+## Reason for decision
+- Modules are a well-known programming language feature that allow for extensibility and separation of concerns.
+- Modules allow distribution implementors and maintainers to extend OpenMRS by using a browser spec, instead of something
+OpenMRS-specific. This means that extending OpenMRS is better understood and more approachable for the frontend community.
+- Modules create better [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), since dependencies
+between modules are explicit and do not rely on global variables.
+- Modules can be lazy loaded, or not loaded at all. This allows OpenMRS distributions to choose which modules they wish
+to include in their code.
+
+## Alternatives
+Instead of using both in-browser modules and build-time modules, we could use build-time modules exclusively. This would
+simplify some things, but would mean that all frontend code would have to go through a single build process, which would be
+very difficult to get working when supporting multiple frameworks and codebases.
+
+Another alternative is to use neither in-browser modules nor build-time modules. This is how much Javascript code is still
+written, including all existing OpenMRS JSPs and GSPs. Not using modules makes it a lot more difficult to build an extensible frontend,
+since you ultimately must rely on load-order and global variables for code to interact with each other.
+Dependencies between global variables are not explicit and harder to manage.
+
+## Common practices (not enforced)
+- When implementing frontend code at OpenMRS, bundle many build-time modules into a single in-browser module by using a javascript bundler.
+- Create only one in-browser module for your application's components, instead of one in-browser module for each component.
--- a/text/0002-modules.md
+++ b/text/0002-modules.md
@@ -15,7 +15,7 @@ exported.
 
 Javascript modules are also known as ECMAScript modules, ES modules, ESM, ES2015 modules, or ES6 modules.
 
-An in-browser javascript module is a javascript module that is loaded in the browser. This is different from
+An in-browser javascript module is a javascript module that is downloaded in the browser. This is different from
 build-time modules, which are compiled down so that they know longer use `import` or `export` by the time they
 are executed in the browser. Most major frameworks (React, Vue, Angular, etc) always use build-time modules, but do not
 always use in-browser modules. At OpenMRS, we use both.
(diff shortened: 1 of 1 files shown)


text/0002-modules.md
```diff
@@ -26,17 +26,18 @@ that there is only one network request for a whole group (bundle) of build-time
 [Browser support for modules](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Browser_compatibility)
 started in Chrome 61, Firefox 60, Edge 16, and Safari 10.1. Internet Explorer does not support modules.
 
-A module loader is a javascript library that allows you to use modules in a browser that does not support them. This makes
-it possible for us to use modules in Internet Explorer.
+A [module loader](https://www.jvandemo.com/a-10-minute-primer-to-javascript-modules-module-formats-module-loaders-and-module-bundlers/) is
+a javascript library that allows you to use modules in a browser that does not support them. This makes it possible for us to use modules
```
dkayiwa: I do not see this as a big deal for us. 😊 
We have the freedom to specify which browsers we do support.
So we can forget supporting internet explorer.

text/0002-modules.md
```diff
@@ -0,0 +1,58 @@
+# Modules
+- Start Date: 2019/04/29
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/2
+
+## Decision, including impact on distributions
+In-browser Javascript modules are the OpenMRS mechanism for adding functionality and creating frontend code for distributions.
```
dkayiwa, line 6: Would this be the equivalent of what we currently have where an end user goes to our [module repository](https://addons.openmrs.org), downloads a module or open web app, and clicks a button in the web application to install it at runtime to extend the the user interface? https://wiki.openmrs.org/x/QQEr
https://wiki.openmrs.org/x/RAEr

text/0002-modules.md
internettrans: Oo really? I have been planning on adding a PR that discusses browser support, but was working under the assumption that we wouldn't have much control over that because of how widespread OpenMRS is being used. Are people currently using IE with OpenMRS? Would removing IE mean that fewer people can use it?

text/0002-modules.md
internettrans: 👍 yes

Both developers and end users could have the ability to pick and choose which frontend javascript modules to include in their application. Developers of distributions would definitely always have this choice, and for end users we would have to write some code and maybe a UI to help them do this, but it is 100% possible with javascipt modules. Sounds like something that we should probably build for end users.

text/0002-modules.md
internettrans: Also, I just pushed a commit that clarifies that using SystemJS is not just for IE11. There are different features of modules and not even Chrome/Firefox/Edge/Safari have one of the important ones yet -- import maps.

dkayiwa: If it is more than just IE, then that is fine.

text/0002-modules.md
```diff
@@ -0,0 +1,59 @@
...
+When browsers fully support modules, the browser will be able to do what SystemJS is currently doing for us.
+
+## Reason for decision
+- Modules are a built-in programming language feature that allows us to easily extend the OpenMRS frontend.
+- Modules allow distribution implementors and maintainers to extend OpenMRS functionality via a browser spec, instead of something
+OpenMRS-specific. This means that the process for extending OpenMRS is better understood and more approachable for frontend developers.
+- Modules create better [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), since dependencies
+between modules are explicit and do not rely on global variables.
+- Modules can be lazy loaded, or not loaded at all. This allows OpenMRS distributions to choose which modules they wish
+to include in their code.
```
mogoodrich, line 45: Great!

text/0002-modules.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Alternatives
+Instead of using both in-browser modules and build-time modules, we could use build-time modules exclusively. This would
+simplify some things, but would mean that all frontend code would have to go through a single build process, which would be
+very difficult to get working when supporting multiple frameworks and codebases.
```
mogoodrich, line 50: Agreed... seems like a maintenance nightmare!

text/0002-modules.md
```diff
@@ -0,0 +1,59 @@
...
+build-time modules, which are compiled down so that they know longer use `import` or `export` by the time they
+are executed in the browser. Most major frameworks (React, Vue, Angular, etc) always use build-time modules, but do not
+always use in-browser modules. At OpenMRS, we use both.
+
+In-browser javascript modules are downloaded with one network request per module. Build-time modules are compiled away such
+that there is only one network request for a whole group (bundle) of build-time modules.
+
+[Basic browser support for modules](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#Browser_compatibility)
+started in Chrome 61, Firefox 60, Edge 16, and Safari 10.1. There are advanced module features, such as import maps, that are not yet
+supported in any browser. Internet Explorer does not have any support for modules.
```
mogoodrich, line 28: FWIW, this is one of the few areas where OpenMRS's use case can make things simpler... it's not an open-to-the-world, customer-facing app, so we can define a minimum spec of browsers to use... I think excluding IE and requiring the browser versions mentioned above is fine.

internettrans: 👍 awesome -- I'm actually going to submit an RFC proposal for browser support next. I'll exclude IE.

text/0002-modules.md
internettrans: 💯

mseaton: @joeldenning, could you give an example of where in-browser modules will simplify things?  Am I understanding right that with build process, your team needs a process to build using webpack and produce a deployable artifact, whereas with the runtime process your team needs a process to maintain an html file that is kind of a runtime equivalent of a webpack config that defines dependencies in script tags, and a main js file to execute?  In both cases, a team needs to write code to create a "distribution" of javascript modules.  I'm trying to understand how a "single build process, which would be very difficult to get working when supporting multiple frameworks and codebases" plays out in practice.  Thanks!

text/0002-modules.md
internettrans: See related #3

text/0002-modules.md
internettrans: > could you give an example of where in-browser modules will simplify things?

This is best explained by talking about what you have to do if you don't have in-browser modules. See my answer below about build process -- hopefully that will answer.

> Am I understanding right that with build process, your team needs a process to build using webpack and produce a deployable artifact,

Webpack is not required (rollup, parcel, jspm also fine), and actually a build process isn't strictly required at all. But yes, most modules would be built with webpack. Webpack is what angular-cli, vue-cli, and react all do by default. I plan on submitting an RFC PR in the next few days about "build process" to clarify exactly what is required.

>  whereas with the runtime process your team needs a process to maintain an html file that is kind of a runtime equivalent of a webpack config that defines dependencies in script tags, and a main js file to execute?

Yeah you have to have a map of all the in-browser modules and their urls. See [import maps](https://docs.google.com/presentation/d/1sv0n_15Zp9HNusdSagOXnBO7Kbh2qURMQj3a8OR8ndo/edit#slide=id.g5681fc88ca_0_146) and [SystemJS](https://github.com/systemjs/systemjs). My next RFC PR will be about Import Maps.

> I'm trying to understand how a "single build process, which would be very difficult to get working when supporting multiple frameworks and codebases" plays out in practice.

Let's break this down into 3 different options for how a large single page application works:

- Option 1: A single webpack build for everything
This would mean that the OpenMRS code and the distribution code would ultimately need to all be in the same file system at build time, and that there would be exactly one webpack config that everything has to live with. If we do not dictate a framework for all OpenMRS (which I do not want to do), a single webpack config takes angular-cli, vue-cli, and create-react-app off the table for people writing OpenMRS code. Reason for that is that each assumes sole and total control over their own webpack config. Those are **the absolute most popular ways to build write frontend code**, so taking them off the table would be a big restriction
- Option 2: Multiple webpack builds, but no in-browser modules
This would mean that create-react-app, angular-cli, and vue-cli are all back on the table. But now how do you get those separate webpack bundles to coordinate in the browser?? Global variables? Communicating through global variables is essentially doing in-browser modules except with implicit dependencies and load-order dependencies.
- Option 3: Multiple webpack builds, with in-browser module
You can use vue-cli, angular-cli, react, or anything else. Plus your code can coordinate/communicate with other modules that aren't even written in the same framework. Modules are universal to javascript.

**Video that will help explain maybe**
For an overview of what this architecture offers and what it could mean for OpenMRS, check out this video I made: https://drive.google.com/file/d/1OkdaDJf4lCTc1FwQcxsWxx-7Kk9WaYpW/view?usp=sharing.

text/0002-modules.md
```diff
@@ -0,0 +1,57 @@
...
+very difficult to get working when supporting multiple frameworks and codebases.
+
+Another alternative is to use neither in-browser modules nor build-time modules. This is how much Javascript code is still
+written, including all existing OpenMRS JSPs and GSPs. Not using modules makes it a lot more difficult to build an extensible frontend,
+since you ultimately must rely on load-order and global variables for code to interact with each other.
+Dependencies between global variables are not explicit and harder to manage.
+
+## Common practices (not enforced)
+- When implementing frontend code at OpenMRS, bundle many build-time modules into a single in-browser module by using a javascript bundler.
+- Create only one in-browser module for your application's components, instead of one in-browser module for each component.
```
fatmali: @joeldenning I have a few questions: 

Since the imports will be done at runtime, I assume end users can have the ability to add more than one in-browser module, will the app take a performance hit when this is done? Can we limit the number of in-browser modules? Is there a possibility that in-browser modules can conflict with each other?

internettrans: > Since the imports will be done at runtime, I assume end users can have the ability to add more than one in-browser module

Yep!

> will the app take a performance hit when this is done?

No. Adding a module means that the javascript code for the module will be downloaded and executed, the same as any other javascript file. But there's no performance overhead beyond just downloading and executing the javascript file.

> Can we limit the number of in-browser modules?

There's not a way of enforcing a limit to # of modules. But ultimately it's not so much about the number of modules as much as the number of bytes of javascript that are being downloaded and executed. 15 modules that are 1 kilobyte is wayyy better for performance than one module that is 1 module that is 15 megabytes. Think of a "module" as just a `<script>` element -- it's not so much how many you have as much as how many bytes they are.

> Is there a possibility that in-browser modules can conflict with each other?

The only unique thing about a module is its name. The code's imports, exports, and variables never conflict. In the following example, you could not add another `@ampath/appointments`, but you could replace it with a different one or create a new module that has some new functionality.

```json
{
  "imports": {
    "@ampath/appointments": "https://ampath.com/appointment.js"
  }
}
```

fatmali:
I approve

brandones:
👍

internettrans:

angshu:
LGTM

internettrans:
Three approvals so I’m merging

PR #3, opened 2019-04-26
Proposal for browser support.

internettrans:

### DIFF
--- a/text/0003-browser-support.md
+++ b/text/0003-browser-support.md
@@ -34,7 +34,7 @@ npm packages.
 ## Reason for decision
 - Not supporting older browers greatly enhances the frontend developer's ability to quickly build quality functionality.
 - Not supporting older browsers allows us to have smaller javascript bundles, which means less code to download and execute in the browser.
-- OpenMRS is usually run by people who can be instructed by modern browsers.
+- OpenMRS is usually used by people who can be instructed by modern browsers.
 
 ## Alternatives
 We could either not define which browsers we support, or we could support older browsers. The former means we wouldn't be certain whether a browser-specific
--- a/text/0003-browser-support.md
+++ b/text/0003-browser-support.md
@@ -1,0 +1,45 @@
+# Project title
+- Start Date: 2019/04/26
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/3
+
+## Decision, including impact on distributions
+Core OpenMRS frontend code supports the following browsers:
+- Last 2 Safari major versions
+- Last 3 Edge major versions
+- Last 5 Chrome major versions
+- Last 3 Firefox major versions
+- Last 3 Opera major versions
+
+The following browsers are not supported:
+- Internet Explorer
+
+Distributions may choose which browsers they support, including to drop support for browsers that they do not care about.
+
+## Definition
+Supporting a browser means ensuring that the javascript code meets the syntax implemented by that browser and does not use any
+DOM or javascript functionality that does not exist or is not implemented in that browser.
+
+Accomplishing this can be done with one of more of the following strategies:
+- Only write code that works all the supported browsers.
+- Write code that is not supported by all browsers, but compile it to older forms of javascript. The most popular options for doing
+this are [babel](https://babeljs.io/) and [typescript](https://www.typescriptlang.org/).
+- Use [polyfills](https://en.wikipedia.org/wiki/Polyfill_(programming)) to enhance older browsers such that they support some
+of the modern javascript features.
+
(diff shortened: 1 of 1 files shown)


text/0003-browser-support.md
```diff
@@ -0,0 +1,45 @@
...
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/3
+
+## Decision, including impact on distributions
+Core OpenMRS frontend code supports the following browsers:
+- Last 2 Safari major versions
+- Last 3 Edge major versions
+- Last 5 Chrome major versions
+- Last 3 Firefox major versions
+- Last 3 Opera major versions
+
```
mseaton, line 12: As context, at PIH we only officially support Chrome.  We don't code or test for any other browser types or versions and make no guarantee about how well things will work.  OpenMRS _may_ want to be more flexible than that, but given limited development and testing resources, I doubt we will every truly certify that all of those listed browsers are supported.

jonathandick: AMPATH only supports chrome as well.

internettrans: The cost of supporting [evergreen browsers](https://www.techopedia.com/definition/31094/evergreen-browser) is very comparable to just supporting chrome. You can support firefox, safari, and edge with literally no work for the majority of code and projects.

I disagree in principle with only supporting chrome -- it sends a ton of usage data to Google servers and is scary if you care about privacy. It's definitely a nice browser and one that should be supported, but I think that saying "you must give all of your browsing history to Google so that they can advertise to you" is too strong of a stance for a large open source project.

My preference is to have OpenMRS support evergreen browsers, and let distributions decide on their own about what they'll support

kaweesi: i have not tried them out but looks like there are some cross-browser testing tools we can use to reduce the testing resource requirements.

internettrans: 👍 yeah i've used saucelabs before and it's alright. Generally speaking, though, much of the code we write will work in evergreen browsers without intensive testing, though.

fatmali:
Approved.

text/0003-browser-support.md
brandones: Big ups to supporting firefox, in light of the recent news about Chrome’s disabling adblockers...

brandones:
👍

internettrans:

text/0003-browser-support.md
```diff
@@ -0,0 +1,45 @@
...
+this are [babel](https://babeljs.io/) and [typescript](https://www.typescriptlang.org/).
+- Use [polyfills](https://en.wikipedia.org/wiki/Polyfill_(programming)) to enhance older browsers such that they support some
+of the modern javascript features.
+
+At OpenMRS, we employ all three strategies, relying most heavily on compilation via Babel. The [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env)
+npm package allows you to choose which browsers you want to support and then will automatically make sure that your modern fancy code works
+in those browsers. It does this by using the [browserslist](https://github.com/browserslist/browserslist) and [compat-table](https://github.com/kangax/compat-table)
+npm packages.
+
+## Reason for decision
```
angshu, line 34: Regarding, supporting older versions of other browsers (current - v2/v3), I would rather accept a bigger download (and hopefully cached so not to download again in defined time), because I am not sure of the assumption of "openmrs is used by people who can be instructed by modern browser"  in all contexts. Many places the browsers at access points, are seldom replaced for years, and auto-update may not work since in-LAN deployment, might not have internet.

angshu:
Ok to merge. See [comment](https://github.com/openmrs/openmrs-rfc-frontend/pull/3/files#r291892405), hope we don't close the door and implementers can choose if they want to support.

text/0003-browser-support.md
dkayiwa: For the implementations that i have been directly involved with, part of the requirements has been to install a certain browser on each client. Could we do a post on talk and ask if there are implementations out there where this would work against them?

internettrans: Yeah thanks for sharing this angshu. Is IE11 support a thing that is a must for a lot of Bahmni users?

internettrans: For reference, here is info about end of life for IE10 and IE11 - https://en.wikipedia.org/wiki/Internet_Explorer_11. Microsoft has announced end of support for IE10 in January 2020, and IE11 will only receive critical security fixes as it becomes older and older.

text/0003-browser-support.md
mogoodrich: I agree with supporting evergreen browsers...

PR #4, opened 2019-05-02
Import maps

internettrans:


text/0004-import-maps.md
```diff
@@ -0,0 +1,82 @@
...
+    <!-- The browser understands the unusual src attribute because of the import map we have defined above -->
+    <script type="module" src="import:@openmrs/styleguide"></script>
+  </head>
+  <body>
+  </body>
+</html>
+```
+
+Because import maps are not implemented in browsers yet, we use [SystemJS](https://github.com/systemjs/systemjs) as a polyfill for import maps.
+The syntax for SystemJS import maps is slightly different than the specification:
```
dkayiwa, line 39: Any reason for making it different from the specification?

internettrans: The only difference is changing the `<script type="importmap">` to `<script type="systemjs-importmap">`. The inner json is exactly the same.

I believe the reason that it is different is so that the developer has full control over whether they are using native browser import maps vs systemjs import maps. Using the exact same script type could result in users accidentally using native browser import maps instead of SystemJS if they update their browser to a version that uses native import maps. Since the specification for import maps is still slightly evolving, it is safer to stick to SystemJS until we're confident that our code will work in all browsers' implementation of import maps.

^^ This is my best guess as to the explanation, but if you want to ask the author of SystemJS, you can open an issue in https://github.com/systemjs/systemjs asking about it. I am a contributor to the project, but not the primary author and maintainer.

dkayiwa: Oh i see! Got you. 😊

fatmali (approved):

brandones (approved):
LGTM

internettrans:

angshu:
LGTM

PR #10, opened 2019-06-09
Proposal: OpenMRS Styleguide and Design Libraries

internettrans:

### DIFF
--- a/text/0010-design-libraries.md
+++ b/text/0010-design-libraries.md
@@ -8,6 +8,8 @@ Material, Semantic UI). Instead, the styleguide is built from scratch in collabo
 UX principles and design elements are liberally borrowed from other design systems and libraries, but the code for the
 OpenMRS styleguide is written by the OpenMRS community.
 
+The decision to embrace one or multiple design libraries for non-code resources will be made separately.
+
 Distributions should use the OpenMRS styleguide as the primary source of making a consistent user experience, look and feel,
 layout, and base-level components. In addition to the OpenMRS styleguide, distributions may use design libraries if they prefer to.
 However, it is not necessary (nor encouraged) to do so.
--- a/text/0010-design-libraries.md
+++ b/text/0010-design-libraries.md
@@ -29,6 +29,12 @@ of the library to make the design library their own.
 - Upgrading design libraries is often very difficult because breaking changes span across an organization's entire
   frontend codebase. Many organizations get stuck on a particular version of a design library and end up having to
   rewrite much of their code to be able to upgrade or switch to a different design library.
+- Backwards compatibility is especially important for the styleguide, since it is shared by everything else. It is
+  difficult to not get stuck on an old version of a design library when upgrading it impacts all frontend code at once.
+  Backwards compatibility is easier to achieve when (1) you have control over the code that needs to be backwards compatible,
+  and (2) you have a minimal API surface (which in this case, means having fewer css classes and javascript components).
+  This is easier to accomplish if we maintain our own. See [this github thread](https://github.com/openmrs/openmrs-rfc-frontend/pull/10#discussion_r292105076)
+  for more discussion.
 - Because of trying to be useful to all organizations in the world, design libraries are often big and bloated. They
   include many features that are unnecessary or even damaging to a user experience when used inappropriately. It is
   difficult for feature developers to know which parts of the design library to use or not.
--- a/text/0010-design-libraries.md
+++ b/text/0010-design-libraries.md
@@ -1,9 +1,9 @@
 # The Styleguide and Design Libraries
 - Start Date: 2019/06/08
-- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/8
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/10
 
 ## Decision, including impact on distributions
-The [OpenMRS styleguide](/text/0006-styleguide.md) does not use a third party design library's code (such as Twitter Bootstrap,
+The [OpenMRS styleguide](/text/0008-styleguide.md) does not use a third party design library's code (such as Twitter Bootstrap,
 Material, Semantic UI). Instead, the styleguide is built from scratch in collaboration with UX designers in the OpenMRS community.
 UX principles and design elements are liberally borrowed from other design systems and libraries, but the code for the
 OpenMRS styleguide is written by the OpenMRS community.
--- a/text/0010-design-libraries.md
+++ b/text/0010-design-libraries.md
@@ -29,6 +29,9 @@ of the library to make the design library their own.
 - Upgrading design libraries is often very difficult because breaking changes span across an organization's entire
   frontend codebase. Many organizations get stuck on a particular version of a design library and end up having to
   rewrite much of their code to be able to upgrade or switch to a different design library.
+- Maintaining our own styleguide enables UX designers in the OpenMRS community. When using a third-party design
+  library, developers often tell a designer "well that's just not possible" or "that's not how <design library> does it."
+  In other words, many design decisions are made by the third party library, and not by designers within the OpenMRS ecosystem.
 - Backwards compatibility is especially important for the styleguide, since it is shared by everything else. It is
   difficult to not get stuck on an old version of a design library when upgrading it impacts all frontend code at once.
   Backwards compatibility is easier to achieve when (1) you have control over the code that needs to be backwards compatible,
--- a/text/0008-design-libraries.md
+++ b/text/0008-design-libraries.md
@@ -3,8 +3,8 @@
 - RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/8
 
 ## Decision, including impact on distributions
-The [OpenMRS styleguide](/text/0006-styleguide.md) does not use a third party design library's code (such as Material,
-Semantic UI, etc.). Instead, the styleguide is built from scratch in collaboration with UX designers in the OpenMRS community.
+The [OpenMRS styleguide](/text/0006-styleguide.md) does not use a third party design library's code (such as Twitter Bootstrap,
+Material, Semantic UI). Instead, the styleguide is built from scratch in collaboration with UX designers in the OpenMRS community.
 UX principles and design elements are liberally borrowed from other design systems and libraries, but the code for the
 OpenMRS styleguide is written by the OpenMRS community.
 
@@ -17,8 +17,8 @@ A design library is a third-party javascript library that provides pre-written c
 are installed into code and are often used by the frontend community for base-level components (such as buttons, form elements,
 and tooltips).
 
-At the time of this writing, [Material Design]() (from Google)
-and [Semantic UI]() are two popular design libraries.
+At the time of this writing, [Material UI](https://material-ui.com/) (from Google), [Twitter Bootstrap](https://getbootstrap.com/),
+and [Semantic UI](https://semantic-ui.com/) are a few of the popular design libraries.
 
 A custom build of a design library is a customized version of such a library. Customizations often include things like custom
 css class naming, additional css class definitions, and custom branding variables. Design libraries often provide a
--- a/text/0008-design-libraries.md
+++ b/text/0008-design-libraries.md
@@ -1,0 +1,47 @@
+# The Styleguide and Design Libraries
+- Start Date: 2019/06/08
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/8
+
+## Decision, including impact on distributions
+The [OpenMRS styleguide](/text/0006-styleguide.md) does not use a third party design library's code (such as Material,
+Semantic UI, etc.). Instead, the styleguide is built from scratch in collaboration with UX designers in the OpenMRS community.
+UX principles and design elements are liberally borrowed from other design systems and libraries, but the code for the
+OpenMRS styleguide is written by the OpenMRS community.
+
+Distributions should use the OpenMRS styleguide as the primary source of making a consistent user experience, look and feel,
+layout, and base-level components. In addition to the OpenMRS styleguide, distributions may use design libraries if they prefer to.
+However, it is not necessary (nor encouraged) to do so.
+
+## Definition
+A design library is a third-party javascript library that provides pre-written css and javascript components. Design libraries
+are installed into code and are often used by the frontend community for base-level components (such as buttons, form elements,
+and tooltips).
+
+At the time of this writing, [Material Design]() (from Google)
+and [Semantic UI]() are two popular design libraries.
+
+A custom build of a design library is a customized version of such a library. Customizations often include things like custom
+css class naming, additional css class definitions, and custom branding variables. Design libraries often provide a
+well-supported way to create a custom build of their library. Organizations can use custom builds instead of the generic version
+of the library to make the design library their own.
+
+## Reason for decision
+- Upgrading design libraries is often very difficult because breaking changes span across an organization's entire
+  frontend codebase. Many organizations get stuck on a particular version of a design library and end up having to
+  rewrite much of their code to be able to upgrade or switch to a different design library.
+- Because of trying to be useful to all organizations in the world, design libraries are often big and bloated. They
+  include many features that are unnecessary or even damaging to a user experience when used inappropriately. It is
+  difficult for feature developers to know which parts of the design library to use or not.
+- Implementing our own styleguide is not intractably difficult. It is something that perhaps the majority of large
+  organizations do. Implementation of a styleguide can be iterative and incremental. Not everything needs to be
+  decided upfront or implemented all at once. It is better to start with something effective and simple than
+  comprehensive.
+- Bug fixing a design library is harder than bug fixing OpenMRS code.
+
+## Alternatives
+The alternative is to use a design library such as Material Design or similar.
+
+## Common practices (not enforced)
+- Distribution code should use the OpenMRS styleguide as much as possible, and should consider not bringing in a design library.
+- OpenMRS developers should contribute to the OpenMRS styleguide when they find bugs or missing features in it. Doing so fosters
+  the health and usability of the styleguide by all distributions.


internettrans:
Will merge if/when there are three approvals from @fatmali @angshu @ojwanganto, @brandones

text/0010-design-libraries.md
```diff
@@ -0,0 +1,47 @@
...
+At the time of this writing, [Material UI](https://material-ui.com/) (from Google), [Twitter Bootstrap](https://getbootstrap.com/),
+and [Semantic UI](https://semantic-ui.com/) are a few of the popular design libraries.
+
+A custom build of a design library is a customized version of such a library. Customizations often include things like custom
+css class naming, additional css class definitions, and custom branding variables. Design libraries often provide a
+well-supported way to create a custom build of their library. Organizations can use custom builds instead of the generic version
+of the library to make the design library their own.
+
+## Reason for decision
+- Upgrading design libraries is often very difficult because breaking changes span across an organization's entire
```
brandones, line 31: Since microfrontends can manage their styling independently, doesn't that enable us to upgrade versions of a library piecemeal across the different components? Maybe some microfrontends will get left behind, but we've explicitly chosen an architecture in which big rewrites are unlikely to be necessary.

text/0010-design-libraries.md
```diff
@@ -0,0 +1,47 @@
...
+  frontend codebase. Many organizations get stuck on a particular version of a design library and end up having to
+  rewrite much of their code to be able to upgrade or switch to a different design library.
+- Because of trying to be useful to all organizations in the world, design libraries are often big and bloated. They
```
brandones, line 43: If the OpenMRS styleguide has a dedicated and judicious owner, perhaps perhaps perhaps we will be able to avoid having something that's big and bloated in addition to being less than state-of-the-art (OpenMRS having somewhat fewer professional engineering and design resources than Google and Twitter).

Another way of saying this, to the bullet point below, is that I'm not sure OpenMRS is a "large organization" in the sense that's meaningful here.

brandones:
While I assume that you have a lot of relevant experience working with organizations that have followed either path (developing their own styleguide vs using a library), and I'm willing to go with your suggestion if your experience points to it being the right one, it goes pretty seriously against my intuition. Sorry to be creating more work for you; I imagine that you're right as usual, but I'm not quite on board yet.

text/0010-design-libraries.md
internettrans: Yes, exactly. The styleguide, though, isn't independent from the other microfrontends, which means that breaking changes to the styleguide impact all of the microfrontends at once.

Here's an example situation that could happen if the styleguide is material-ui:
1) Styleguide uses material-ui@4.1.0, since that's the current latest version.
2) All microfrontends use the css classes exposed by material-ui
3) In 1-2 years, material-ui@5 comes out. It has some breaking changes to the exposed css classes that impact some, but not all, microfrontends.
4) What do we do with the styleguide?
a) We decide not to upgrade styleguide to material-ui@5 until all the important microfrontends are forwards-compatible with material-ui@5. It is logged as technical debt.
b) OR: we decide to upgrade styleguide to material-ui@5 immediately, encouraging all microfrontends to upgrade as soon as possible.

I don't like either of the outcomes very much. 4a often results in the styleguide remaining on material-ui@4 for a very long time. Compare this to OpenMRS recently getting to Java 8 after what I understand to have been a significant effort, even though Java 12 is the current version. Or another one is how much of OpenMRS frontend uses jquery@1 (at least the few files i've looked at), even though the latest version is jquery 3.

For 4b, I've seen immediate upgrades to a core library like that result in people staying on older versions of the core library indefinitely because they never get around to material-ui forwards compatibility. For example, my understanding is that Kenya EMR is still on OpenMRS@1 because upgrading would be hard.

^^ All of this just goes to show the importance of backwards compatibility with shared, core code. The more that code is reused and shared, the more hesitant we should be to make breaking changes to it. And I believe backwards compatibility to be far easier to accomplish if you have 1) a minimal API surface (in this case, less css classes and components) and 2) control over the code that needs to be backwards compatible.

brandones: Oh, wow, I understand now. Yep.

text/0010-design-libraries.md
internettrans: Agreed.

The point I make in the bullet point is referring to if we adopted an extant third party design library, such as Bootstrap, that we would be inheriting their bloat.

internettrans:
@brandones what does your intuition tell you? Would you embrace something like material-ui,  semantic-ui, or bootstrap for creating a styleguide?

text/0010-design-libraries.md
brandones: But wait, wouldn't our library be subject to the same constraints? Is there a big difference between not being allowed to upgrade from Material UI 4 to Material UI 5, and not being allowed to make breaking changes in our own library?

internettrans: > Is there a big difference between not being allowed to upgrade from Material UI 4 to Material UI 5, and not being allowed to make breaking changes in our own library?

Nope. No difference at all. A breaking change to a homegrown styleguide is just as bad as a breaking change to a third party styleguide. Which is why I think we should keep core code as backwards compatible as possible. And that writing and maintaining that code ourselves is the best way to achieve that.

From my earlier comment:
> I believe backwards compatibility to be far easier to accomplish if you have 1) a minimal API surface (in this case, less css classes and components) and 2) control over the code that needs to be backwards compatible.

brandones:
My intuition is to avoid solving solved problems and avoid introducing more code into the world. Big teams of very talented engineers and designers have worked on this for 7+ years now. It seems like a hard sell to decide to do it again, in-house, given the resources we have available to dedicate to it.

If this were my pet project, I'd use Bootstrap because that's what I'm familiar with, but could easily be swayed.

text/0010-design-libraries.md
internettrans: I'm going to update the PR to add a bullet point about this ^^ --  maintaining backwards compatibility wherever possible in the styleguide is important, and writing our own code facilitates that imo.

brandones: Nice. Ok, that's quite compelling.

internettrans:
Yeah, reinventing the wheel is never fun. Consider this example, though, of a styleguide: https://storybook.jumpsoftware.com/. It was built from scratch by one developer (in collaboration with a designer) in a couple weeks. It is being added to incrementally and isn't fully featured yet, but it has the things that are needed so far for Jump Software. The things added to the styleguide are done so in collaboration with the designers at Jump, and are applied company-wide. I helped a bit with this styleguide (mostly through code review, but also wrote some of it myself) and the task was not astronomical.

text/0010-design-libraries.md
dkayiwa: In the OpenMRS community, there has been a general drive to run away from our custom things that we have to maintain, force new comers to learn, to get towards such things that new developers are already familiar with from other projects and which we do not have to maintain ourselves. Should we bring this to a wider audience on talk to get what others think? https://wiki.openmrs.org/display/RES/2017-01-23+Design+Forum

internettrans: Could you explain your reservations more, @dkayiwa? Do you prefer using a customized build of Twitter Bootstrap / similar? Or something else entirely?

internettrans:
I created a twitter poll for anybody interested in reading some broader discussion on the subject: https://twitter.com/Joelbdenning/status/1138169008279248901

text/0010-design-libraries.md
dkayiwa: This was not me. Was mainly re-echoing what i have seen discussed.
cc @ssmusoke @bmamlin

mogoodrich:
Thanks for all the discussions here! Agree with Daniel thiat a summary Talk post would be worthwhile for broader discussion.

internettrans:
@mogoodrich 👍 JJ just posted in Talk a link to this and the other styleguide RFCs. https://talk.openmrs.org/t/micro-frontends-architecture-for-openmrs/22557/24. Hopefully that will help drive people to this PR and be able to comment and add to the discussion

dkayiwa:
@joeldenning  The OpenMRS community has a category of people who would love to help, but do not have the time to follow all these pull requests and their entire discussion threads or comments. So just pointing them to certain parts where we have differing views, may be a good starting point. Of course they can as well look at the rest, if they have the time. 😊

mseaton:
FYI - I do think this is an area that might require further debate and discussion.  I think there are strong arguments to be made on both sides.

gregoryjschmidt:
Hi 
_In short_, I was surprised when Joel suggested we build our own 'atomic' components from scratch (rather than use an existing design library), but this does seem to make sense to me now. 

This RFC's proposed route provides a 'best of both world's approach. (Custom code, but use established design principles).

**A) Build the components 'from scratch'** 
a1) because we don't inherit bloat of larger libraries 
a2) because we have more control
a3) we can have a more contemporary design (see below)
a4) we can design these components with their anticipated compilation in larger widgets (see below)

**B) Strongly base the user interaction experience of those designs upon existing 'design languages' and 'established UI design principles'** such as Google Material Design, and Android Design.
b1) to take full benefit of the research and familiarity users have with standard existing user interaction design principles and styling that people are accustomed to

---
_Longer answer to a3/a4:_
The original mockups I was making used stock Google Material Design components as the atomic elements, and then I was focusing on using these pieces to make the larger 'molecules' / 'widgets'. 

But one problem I had with the current libraries for GMD is they actually seem rather out of date relative to the current styling of Google's Material Design after their app re-fresh over the last year. I was hoping for a massive update to all the GMD components styling at Google I/O in early May this year, but the 'new version' never came. The only way we can achieve the more contemporary styling is to build these atomic elements ourselves. 

I also think we will have a unique use case for these atomic elements. Although many organizations focus on their atomic element libraries, I think in addition to that we will require lot of effort on the larger 'widget components' (eg. re-usable lab panel, or re-usable medication list panel); and it may be much easier to build these larger re-usable widget components if we have more control over how the smaller atomic elements are coded.

There also can be some cool things we can do with accessibility (size / spacing / colors), internationalization (colors), and branding (colors) if we have more granular control over the code of the atomic elements.

---
Language clarification. Use of term 'atomic' here is similar to how Brad Frost uses 'atomic'. My use of the world 'widget' likely matches closer to Brad's use of the term 'Organism'. Read further [Atomic Design](http://bradfrost.com/blog/post/atomic-web-design/)

brandones:
In my usual frontend development, I rely very heavily on Bootstrap grids. Maybe that makes me a fuddy-duddy. Is the modern approach just to use flexbox directly?

internettrans:
@brandones browsers now have native support for [CSS grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout). I personally use flexbox much more than CSS grid, but both of them are great options for day-to-day layout of things.

Also, we are going to be holding a design forum for this topic, as suggested by @dkayiwa and @mseaton. Cynthia will be helping us set up the  call. Proposed date is 2019/06/24

isaacholeman:
Hi Everyone,

Enjoying catching up on the discussion! I wanted to offer some reflections from a UX perspective. As a designer, I tend to think of Bootstrap, Semanti UI, and Material UI as UI component libraries. In a component library, the code _is_ the library. UI component libraries don’t just include, they really consist of CSS classes, CSS variables, and js components. 

I would distinguish this with what I have in mind when I talk about design libraries with other designers. For example Google’s [Material Design](https://material.io/design/), Microsoft’s [Fluent and Inclusive Design systems](https://www.microsoft.com/design/inclusive/), and [Apple’s Human Interface Guidelines](https://developer.apple.com/design/) emphasize UX guidance and best practices, icon libraries, sketch toolkits, and other great stuff for designers. These resouces establish a consistent/familiar design language, style, or vision; code snippets are offered to help people execute that vision in their custom projects. 

From a design perspective, the decision about whether or not to embrace a specific design library/system is important for two reasons:
1. The less in-house UX capacity you have, the more value there is in accepting the guidance of experts who have had time to think about issues that you don’t have time to properly investigate.
2. Documenting design decisions is hard, and many designers find it tedious (just like code docs!). If the OpenMRS community decides to embrace the guidance of a specific design library, we can say e.g., “for more explanation or guidance on how to customize this for your own purposes, see this [link] page on the xxx design system website.” Without good design documentation, 1-3 UX designers might borrow liberally from a range of design systems without explaining their work, and any non-expert who doesn’t have time to investage all the available design systems just has to take the design experts’ word for it, which is less empowering than if there’s one set of design guidance/documentation that more people could have time to review independently. 

My apologies if these considerations seem tangential to the general thrust of this discussion! To draw this back in, here are my two questions for the community:
1. I would propose to clarify that this PR concerns the use of a third party design library’s code, and that the decision to embrace one or multiple design libraries for non-code resources will be made separately. How do others feel about that?
2. Should we create another PR focused on whether/how to use one or more design libraries for OpenMRS UX work?

Also, I’m going to be at the first day of the hackathon next week, looking forward to discussing further in person!

internettrans:
Hey @isaacholeman, thanks for the in-depth comments. Yeah the terminology here of "design library," "design system," "component library," and "styleguide" are good ones to try to clarify and define. What you said is definitely true -- how a designer talks about using a design system is very different than how a developer might talk about it. One is talking about principles, practices, and methodologies, while the other is talking about code.

The decision about whether/which/how to borrow from or fully adopt a design library / design system / component library is something that should be made together as a group of devs and designers, since it impacts both groups tremendously.

> I would propose to clarify that this PR concerns the use of a third party design library’s code, and that the decision to embrace one or multiple design libraries for non-code resources will be made separately. How do others feel about that?

👍 I have [updated this PR](https://github.com/openmrs/openmrs-rfc-frontend/pull/10/commits/3e4d88593dd2b573e589effa3d655341788916ee) to clarify this. The first two paragraphs of [the Decision section](https://github.com/openmrs/openmrs-rfc-frontend/pull/10/files#diff-ebe1b691d047bcad29eaecb7fe13b010R5) hopefully are now clear about it. I completely agree that our decision about what we use to make our UX Designs is a separate decision.

> Should we create another PR focused on whether/how to use one or more design libraries for OpenMRS UX work?

I would love that! And would love if you, @gregoryjschmidt, or other interested designers led the charge on the proposal, even authoring it if possible (here's [a template](https://github.com/openmrs/openmrs-rfc-frontend/blob/master/rfc-template.md)). Making decisions as a community regarding how things will be designed, user tested, validated, etc would be amazing. And I think it would be great if we formed a "microfrontends UX designer squad" that leads the effort right now in collaboration with the current devs on the microfrontends squad. Or maybe it's all just the same squad and it's just designers joining the current squad calls and helping create a pipeline of work for us.

We are getting close on the code side to being ready to tackle some feature development. And so it would help us a lot to figure out how we get a functioning collaboration and pipeline of work with UX Designers.

isaacholeman:
Thanks @joeldenning, the changes look good! I hadn't thought about the second bullet point in the Reasons for Decision section (devs may tell designers, "that's not how it works"), but it rings true to my experience :). 

The question of whether to adopt or build in-house the code for the style guide is beyond my technical depth, I'll mostly leave that to the devs. I will say though that I'd highly value decisions that increase possibilities for incremental improvement. This is both because of the reason @gregoryjschmidt mentioned above around design patterns changing faster/differently than releases of design libraries. It's also because it's hard to get major dedicated funding for UX work in this space, so it's valuable to be able to build smaller incremental UI improvements within funding streams that are mostly focused on a specific implementation or new feature. But whether this outweighs the value of adopting components that might enable a faster start is a decision for which I'll trust the community's technical judgement.

I've also been in touch with @gregoryjschmidt about this offline. I'm excited to continue the conversation about a design squad and how to start feeding the devs a practical and well-prioritized pipeline of work. We'll work on drafting that new proposal!

cromwel:
Following these conversation and the Micro-front-end discussions i believe Material-UI is the proposed UI framework being used, right?

internettrans:
@cromwel, Material-UI is not what is proposed in this RFC. See [this section](https://github.com/openmrs/openmrs-rfc-frontend/blob/3e4d88593dd2b573e589effa3d655341788916ee/text/0010-design-libraries.md#decision-including-impact-on-distributions) for what is in the current proposal. We are having a design forum Monday to discuss the options, though. Please join!

cromwel:
Thanks @joeldenning it's clear now. The recommended one is the OMRS style guide but more options will be discussed. i'll definitely join the call.

gregoryjschmidt:
In summary: 
- OpenMRS will write code for own style guide / design system. With the intent it will be less bloated, and easier to maintain/upgraded than using a third party design system.
- The OpenMRS style guide / design system will be visually based off standard user interface principles found in mainstream systems (such as Google Material Design).
- If OpenMRS is unsuccessful at maintaining code of its own design library, then it will consider adopting the codebase of an existing design system.

Do we have consensus to merge? It looked like many groups were willing to try this.

internettrans:
Another consideration -- it is [not possible to have a single version of material ui across frameworks](https://twitter.com/MaterialUI/status/1149720000669868033)

dkayiwa:
I have no objection trying that out and see how it goes. 😊

brandones (approved):

mogoodrich:
Works for me!

isaacholeman:
Cool, I’m excited to see this go forward!

internettrans:
For anyone interested, here's the code repo for the styleguide. https://github.com/openmrs/openmrs-esm-styleguide

fatmali:
I approve.

internettrans:
That's two approvals from the MF squad -- one more and we'll merge.

text/0010-design-libraries.md
angshu: I can understand the reluctance given very few UX experts for home-grown. But I agree with Joel on dependency mess and imposition/tight-binding due to 3rd parties! If we keep the surface lean, css classes for essential/base elements, and if that can be overridden by css vars, then I think in the long run it will benefit all. However, I hope this does not create bottlenecks on experts. 

Also even with OpenMRS custom style lib, it does not stop others from using their own or mix with other libraries. If we document "how to override" the documented base styles of OpenMRS css libs (hopefully a small footprint), then this would be easier to maintain/override as well, in case of breaking changes of OpenMRS css lib or specific-upgrades of 3rd party libs.

internettrans:
We are moving forward with this approach in https://github.com/openmrs/openmrs-esm-styleguide and https://openmrs.github.io/openmrs-esm-styleguide/?path=/story/*.

We'll revisit if it doesn't work out, but I'm going to merge this PR since it represents what the MF squad is moving forward with right now.

PR #11, opened 2019-06-17
Proposal: Namespaces

ssmusoke:
How to name the modules

### DIFF
--- a/text/0011-namespaces.md
+++ b/text/0011-namespaces.md
@@ -1,0 +1,29 @@
+# Namespaces 
+- Start Date: 2019/06/17
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/11
+
+## Base Namespace 
+The base namespace for maps and modules will be `@openmrs`
+
+## Core Features 
+The namespace for core features will be `@openmrs/core` which will include core atoms 
+
+## Module based namespaces 
+The namespace for module based features, which will leverage existing modules will be `@openmrs/module/modulename` for example 
+- `@openmr/module/appointmentscheduling` for appointment scheduling 
+
+## Implementing Partner and Distributions 
+The namespace for partners will be as follows
+- `@openmrs/reference-application/` for the Reference Application 
+- `@openmrs/ugandaemr/` for UgandaEMR 
+- `@openmrs/ampath/` for Ampath specific needs 
+- `@openmrs/hiv` for an HIV based distribution 
+
+## Reason for decision
+- To provide a guide for naming modules to ease the process of discoverability  
+
+## Alternatives
+- N/A 
+
+## Common practices (not enforced)
+- NPM package naming - https://docs.npmjs.com/files/package.json#name
--- a/text/0011-namespaces.md
+++ b/text/0011-namespaces.md
@@ -13,14 +13,15 @@ The namespace for module based features, which will leverage existing modules wi
 - `@openmr/module/appointmentscheduling` for appointment scheduling 
 
 ## Implementing Partner and Distributions 
-The namespace for partners will be as follows
-- `@openmrs/reference-application/` for the Reference Application 
-- `@openmrs/ugandaemr/` for UgandaEMR 
-- `@openmrs/ampath/` for Ampath specific needs 
-- `@openmrs/hiv` for an HIV based distribution 
+The namespace for distributions and partners will be as follows
+- `@openmrs/distro/reference-application/` for the Reference Application 
+- `@openmrs/distro/ugandaemr/` for UgandaEMR
+- `@openmrs/distro/hiv` for an HIV based distribution (does not exist - but if one was to be built)
+- `@openmrs/ampath/` for Ampath specific needs. The name would map to an already owned domain name for clarity
 
 ## Reason for decision
 - To provide a guide for naming modules to ease the process of discoverability  
+- The premise is that there will also be very 
 
 ## Alternatives
 - N/A 


text/0011-namespaces.md
```diff
@@ -0,0 +1,29 @@
...
+## Module based namespaces 
+The namespace for module based features, which will leverage existing modules will be `@openmrs/module/modulename` for example 
+- `@openmr/module/appointmentscheduling` for appointment scheduling 
+
+## Implementing Partner and Distributions 
+The namespace for partners will be as follows
+- `@openmrs/reference-application/` for the Reference Application 
+- `@openmrs/ugandaemr/` for UgandaEMR 
+- `@openmrs/ampath/` for Ampath specific needs 
+- `@openmrs/hiv` for an HIV based distribution
```
bmamlin: This last example seems rife for conflicts. The top level path should identify the "authority" responsible for anything within that path & below:
* `module` – module authors are responsible for using a unique module name
* `reference-application` – the team working on the OpenMRS Reference Application
* `ampath`, `pih`, etc. – the organization (ideally would align with owned domain name)
* `bahmni`, `ugandaemr`, etc. – the distribution (ideally would align with owned domain name)
* `hiv` – **who owns this namespace?**

ssmusoke: @bmamlin Thanks for the review, thinking now of a distro prefix for distributions and an organization at the root - this provides segregation at that level

text/0011-namespaces.md
```diff
@@ -0,0 +1,30 @@
...
+## Core Features 
+The namespace for core features will be `@openmrs/core` which will include core atoms 
+
+## Module based namespaces 
+The namespace for module based features, which will leverage existing modules will be `@openmrs/module/modulename` for example 
+- `@openmr/module/appointmentscheduling` for appointment scheduling 
+
+## Implementing Partner and Distributions 
+The namespace for distributions and partners will be as follows
+- `@openmrs/distro/reference-application/` for the Reference Application
```
brandones, line 17: `@openmrs/distro/refapp/`?

ssmusoke: refapp is just a common "nick name", but reference-application is the name of the module and the distribution too. More text to type but more consistent usage across the board 

I am happy to remove the "-", just that I think it aids readability

brandones:
Should we namespace in this way? It would seem to encourage people to build less reusable components.

I actually really like that all `openmrs` stuff is in `openmrs-module-`. It feels weird to have things called `openmrs-module-mysitewidget` and `openmrs-module-mysitething`, and I think that's good. I think that, by implying in our naming conventions that everything should be reusable, we actually encourage people to make more things that are reusable.

If someone wants to make `@openmrs/pih-bmi-graph`, we should do our best to make them feel uncomfortable about it. I think not letting them call it `@openmrs/distro/pih/bmi-graph` would help.

ssmusoke:
@brandones I agree with the sentiment, however the practicality is that there may be very specific functionality that is needed by distributions which may not be generic enough or even useable in the openmrs context at the time.

So the better approach is to explicitly provide a meaningful way of naming for those scenarios than leave it open.

brandones:
Can't they continue to do what they presently do -- create things like `openmrs-module-mirebalaisreports` (but here, `@openmrs/mirebalaisreports`)?

Are frontends likely to need to be more site-specific than modules?

ssmusoke:
Oh yes I expect the front ends to be more site specific and have a higher rate of change than modules - even in UgandaEMR, where we have a baseline distribution module, I expect this will lead to more "site" specific front ends, such as point-of-care, retrospective-entry, regional-referral (larger sites) etc

brandones:
Why don't you build these things general but configurable, and configure to meet your sites' needs? My site in Mexico has POC, retrospective entry, and referral to higher levels of care.

ssmusoke:
Oh yes they can be built and evolved but over time, not in a single shot, hence the need to provide an explicit way of having that support for adoption at site and implementation level

mogoodrich:
@brandones brings up an interesting point... I haven't read the whole spec, but could we create a namespace like "common". ie "@openmrs/common/xyz"?

Also, tangentially and kind of thinking out loud, but I could think of doing something like "@openmrs/pih/common/*" where PIH would put things they "own" but designed to be used by others...

text/0011-namespaces.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Base Namespace 
+The base namespace for maps and modules will be `@openmrs`
+
+## Core Features 
+The namespace for core features will be `@openmrs/core` which will include core atoms 
+
+## Module based namespaces 
+The namespace for module based features, which will leverage existing modules will be `@openmrs/module/modulename` for example 
+- `@openmr/module/appointmentscheduling` for appointment scheduling
```
internettrans, line 13: typo: `@openmrs`

text/0011-namespaces.md
```diff
@@ -0,0 +1,30 @@
...
+## Module based namespaces 
+The namespace for module based features, which will leverage existing modules will be `@openmrs/module/modulename` for example 
+- `@openmr/module/appointmentscheduling` for appointment scheduling 
+
+## Implementing Partner and Distributions 
+The namespace for distributions and partners will be as follows
+- `@openmrs/distro/reference-application/` for the Reference Application 
+- `@openmrs/distro/ugandaemr/` for UgandaEMR
+- `@openmrs/distro/hiv` for an HIV based distribution (does not exist - but if one was to be built)
+- `@openmrs/ampath/` for Ampath specific needs. The name would map to an already owned domain name for clarity
```
internettrans, line 20: why is this one not under `@openmrs/distro`? Also, what do you mean by "The name would map to an already owned domain name for clarity"?

text/0011-namespaces.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Reason for decision
+- To provide a guide for naming modules to ease the process of discoverability  
+- The premise is that there will also be very
```
internettrans, line 24: This sentence is incomplete

text/0011-namespaces.md
internettrans: Hmm, this is making me think (in a good way) about what this new reference application would be. The way that I have  been thinking about the new microfrontends architecture is more that the default implementation of each of the core modules would together comprise the equivalent of a ref app. And that the ref app itself wouldn't be a module, but just the set of all core modules.

That line of thinking is shown in [this line of code in openmrs-esm-root-config](https://github.com/openmrs/openmrs-esm-root-config/blob/master/src/root-config-lib.js#L10), where a function is exported that registers *all* the core applications. Also see [this line](https://github.com/openmrs/openmrs-esm-root-config/blob/master/src/root-config-lib.js#L20) and [this file](https://github.com/openmrs/openmrs-esm-root-config/blob/master/src/root-config-dist.js), which try to make it easy for devs who want the default modules and also for those who do not.

What are your thoughts on this? Is there a specific need for a module called ref app, instead of it just being the collective set of core modules?

internettrans (commented):
I'm glad you took the initiative to create this PR. Naming these modules in a consistent way will provide devs a much easier way of discovering and understanding the frontend modules available to them.

internettrans (commented):
@ssmusoke @brandones @bmamlin @mogoodrich 

How should we proceed on this PR? I don't want it to die -- this is an important topic to iron out. There have been lots of ideas shared. Can everyone share their opinions on the following four topics?

1) Module names in an import map [do not have to follow any rules](https://github.com/WICG/import-maps/issues/63). I propose that we impose on OpenMRS modules the [rules npm has for naming modules](https://docs.npmjs.com/files/package.json#name). The reason for this is that the npm rules are well understood and the industry standard. Additionally, we'll need to publish our modules to npm to facilitate reuse and I think it would be far easier to only have one name that is shared between both the import map and npmjs.com.

2) The `@orgname/packagename` syntax is [explicitly explained in the "npm scopes" documentation](https://docs.npmjs.com/misc/scope). Since we plan to publish these modules to npm (in addition to being in the import map), I propose that we put the name of the [owning npm organization]() as the scope (`@orgname`). Since not every developer will have publish permissions to the `@openmrs` organization on npm, I suggest that we don't put distro-specific things under `@openmrs`, since they're maintained by the community instead of OpenMRS. For example, PIH would scope their modules under `@pih/` on NPM.

3) Slash characters (`/`) have special meaning in npm packages. They refer to importing a javascript file within **a subdirectory of the npm package**. The only exception to this is when the `/` character separates the "scope" from "package name" (see [explanation](https://docs.npmjs.com/misc/scope)). I propose that we start with (and maybe stick to) only importing the main entry file for the package, instead of looking at subdirectories. For this reason, I suggest we do not use the slash character `/` except as specified by the [npm scopes explanation](https://docs.npmjs.com/misc/scope).

4) I agree with Brandon's sentiment that the java naming convention of `openmrs-module-somename` has been successful, and have copied that with naming our current frontend modules with `openmrs-esm-login` and similar. A question is how these should be published to npm and referenced in the import map. One option is `@openmrs/esm-login`. Another is `@openmrs/login`, with the `esm` being assumed since all of the modules are ES modules. I lean towards `@openmrs/login` (without the esm) as the name in the npm package and import map. Thoughts?

mogoodrich:
Thanks @joeldenning .... I was away last week, just getting to this now...

I liked the idea of sub-namespacing via slashes but after reading your point about NPM standards I agree that we should be as consistent as possible.

However, that really only leaves us with a one-level hierarchy, which doesn't allow us a ton of flexibility...

Are we sure that all the modules published to NPM will be "esm"?  I'm actually leaning now toward the a more verbose approach:

@openmrs/esm-login

or perhaps even:

@openmrs/openmrs-esm-login

Reasoning... I'm thinking a little deeper about how other organizations will scope their packages.  I think "openmrs" should be included in the fully scoped name one way or another, as an organization's work may not be limited to OpenMRS... so "@pih/esm-login" or "@pih/login" may not be descriptive enough.

Are there any restrictions/conventions on defining scopes?  Thoughts on whether we'd want to do:

@pih/openmrs-esm-login
@pih/someotherapp-esm-login

or

@pih-openmrs/esm-login
@pih-someotherapp/esm-login

How we decide to go with this may affect how we want to define elements under the @openmrs scope.

Or maybe I'm making this more complicated than it needed to be.  :)

For reference, here are the current things in NPM under @openmrs:

https://www.npmjs.com/search?q=%40openmrs

Also flagging @mseaton  for his thoughts.

internettrans:
@ssmusoke are you interested in continuing to push this RFC forward?

If so, I've got some updates to share with you to help get this in a mergeable state.

internettrans:
Closing this because of inactivity

PR #12, opened 2019-07-25
Proposal: Styleguide Javascript Components

internettrans:
This RFC outlines the usage of javascript components in the OpenMRS styleguide. It will be merged if it gains three approvals from members of the microfrontends squad, as outlined in the RFC's readme.

### DIFF
--- a/text/0012-styleguide-javascript-components.md
+++ b/text/0012-styleguide-javascript-components.md
@@ -5,8 +5,8 @@
 
 ## Decision, including impact on distributions
 The [OpenMRS styleguide](/text/0008-styleguide.md) provides a CSS-first API to be used by all
-frontend code at OpenMRS. Framework-specific Javascript components (for React, Vue, and/or Angular) will be
-implemented when convenient, but are not required for using the styleguide's global css classes.
+frontend code at OpenMRS. For the few elements for which JavaScript components are required (e.g. multiselect),
+the Styleguide will provide framework-specific components in React, Vue, Angular, and/or other frameworks.
 
 ## Definition
 "Javascript component" refers to javascript code that encapsulates the logic for creating and updating
--- a/text/0012-styleguide-javascript-components.md
+++ b/text/0012-styleguide-javascript-components.md
@@ -1,0 +1,64 @@
+# Styleguide javascript components
+- Start Date: 2019/07/24
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/12
+- Previous, defunct RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/9
+
+## Decision, including impact on distributions
+The [OpenMRS styleguide](/text/0008-styleguide.md) provides a CSS-first API to be used by all
+frontend code at OpenMRS. Framework-specific Javascript components (for React, Vue, and/or Angular) will be
+implemented when convenient, but are not required for using the styleguide's global css classes.
+
+## Definition
+"Javascript component" refers to javascript code that encapsulates the logic for creating and updating
+DOM elements and css styles. Components can be reused and composed, but are usually specific to a particular framework.
+
+"Framework agnosticism" refers to implementing the styleguide such that it can be used by many or all frameworks. This means
+that React, Angular, Vue, Svelte, Elm, etc. all can use the component.
+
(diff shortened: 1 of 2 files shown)


text/0012-styleguide-javascript-components.md
```diff
@@ -0,0 +1,64 @@
...
+  expect all global css classes to have a javascript wrapper component in each of the popular javascript frameworks. If the component exists,
+  that's great, but the CSS is the Styleguide API and the thing to rely on.
+- By not betting entirely on the future of any single framework, we are more likely to avoid full rewrites while being able to do
+  incremental rewrites. Web components in the styleguide provide a clear path for migrating a module from one framework to another.
+  Experimentation in new technologies and frameworks is practical only if the experiments don't have to re-implement the entire styleguide
+  just to get going.
+- Web components or other framework agnostic javascript components are
+  [very difficult to implement well](https://medium.com/canopy-tax/one-companys-relationship-with-custom-elements-d360baf3b253) and
+  have [some major drawbacks](https://dev.to/richharris/why-i-don-t-use-web-components-2cia). Implementing simple components, such as
+  a styleguide button, can become [quite complicated](https://twitter.com/Joelbdenning/status/1149822754016780288?s=20) when done with web components.
```
brandones, line 36: Until reading that thread, it had been years since I'd felt like I really didn't understand front-end development.

brandones:
Update the link to "Styleguide Components RFC" in https://github.com/openmrs/openmrs-rfc-frontend/blob/master/text/0008-styleguide.md as well?

gregoryjschmidt:
Thanks Joel, this looks pretty straightforward next step

internettrans:
@brandones 👍 good catch - I have updated that link

internettrans:
@brandones any other comments?

text/0012-styleguide-javascript-components.md
```diff
@@ -0,0 +1,64 @@
+# Styleguide javascript components
+- Start Date: 2019/07/24
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/12
+- Previous, defunct RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/9
+
+## Decision, including impact on distributions
+The [OpenMRS styleguide](/text/0008-styleguide.md) provides a CSS-first API to be used by all
+frontend code at OpenMRS. Framework-specific Javascript components (for React, Vue, and/or Angular) will be
+implemented when convenient, but are not required for using the styleguide's global css classes.
```
brandones: I guess I'm not totally sure what this means. "Will be implemented when convenient" means that Styleguide managers will implement these components for certain specific elements? Or means that people are welcome to implement components, but that doing so will not be necessary to use the OpenMRS Styleguide? (That "[javascript components] are not required for using the styleguide's global css classes" seems like it should be a truism, but maybe there are libraries that have global CSS classes that are only usable with javascript components?)

internettrans: Yeah good question -- what I'm trying to emphasize is that the styleguide's primary API is the css, not the js components. The reason for this is that making cross-framework js components for a styleguide is quite difficult.

Most of the styleguide won't have any js components at all. But for some things within the styleguide, such as multiselects, we will need js components. What this bullet point in the RFC is trying to say is that we can't commit to having a js component in every possible framework for multiselects. We will build the js components only as needed in each of the frameworks.

gregoryjschmidt:
I never realized how much was possible in pure CSS.

Here's an example of a library that shows how standard complex Google Material Design components can be built in pure CSS. 

https://github.com/finnhvman/matter

https://codepen.io/finnhvman/full/zMKagM

text/0012-styleguide-javascript-components.md
brandones: How about

"For the few elements for which JavaScript components are required (e.g. multiselect), the Styleguide will provide a specification (and a reference implementation?). Implementations of that specification for other frameworks will be integrated into the Styleguide whenever they happen to be written."

internettrans: 👍 updated this based on your feedback.

internettrans:
Merging this because this is the direction that we are going, even though we haven't been able to drive the conversation here in this Github PR.

PR #13, opened 2019-08-02
Proposal: API Calls and FHIR

internettrans:
This proposes a way of making api calls from browser to server.

### DIFF
--- a/text/0013-api-calls-and-fhir.md
+++ b/text/0013-api-calls-and-fhir.md
@@ -1,0 +1,76 @@
+# API Calls and FHIR
+- Start Date: 2019/07/02
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/13
+
+## Decision, including impact on distributions
+All API calls from the browser to the server go through the `@openmrs/api` javascript
+module.
+
+The `@openmrs/api` provides the following:
+- `authenticatedFetch`
+- `fhir`
+- `getCurrentPatient`, and a few similar `getCurrent...` functions for shared data.
+
+Whenever a FHIR backend API is available, it should be used. When no such FHIR api exists,
+the OpenMRS-specific APIs will be used.
+
+## Definition
+- `fhir` refers to the [official fhir.js npm package](https://github.com/FHIR/fhir.js/). [FHIR itself](https://www.hl7.org/fhir/overview.html)
+  is an acronym that refers to the industry standard API specification for healthcare data. At the time of this writing, the OpenMRS
+  backend largely offers only non-FHIR APIs for the frontend to use.
+- Within a browser, an API call is an HTTP request to the server. This can be done with either
+  [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) or
+  [XMLHttpRequest XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest).
+  Fetch is preferred when possible, since it is newer and easier to work with (uses Promises).
+- Authentication of an API call is accomplished through a [`Cookie` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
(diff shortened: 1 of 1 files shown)


text/0013-api-calls-and-fhir.md
```diff
@@ -0,0 +1,76 @@
...
+The `@openmrs/api` provides the following:
+- `authenticatedFetch`
+- `fhir`
+- `getCurrentPatient`, and a few similar `getCurrent...` functions for shared data.
+
+Whenever a FHIR backend API is available, it should be used. When no such FHIR api exists,
+the OpenMRS-specific APIs will be used.
+
+## Definition
+- `fhir` refers to the [official fhir.js npm package](https://github.com/FHIR/fhir.js/). [FHIR itself](https://www.hl7.org/fhir/overview.html)
```
gregoryjschmidt, line 18: Have we considered using the java based HAPI FHIR APIs by James Agnew & team, https://github.com/jamesagnew/hapi-fhir
Their project seems to be more active than the fhir.js one

angshu: yes. We should evaluate which library to use. However, cautioning again on the common model, which i am sure will be difficult to agree to! (e.g. Patient attributes, would probably need specific extensions, or relationships etc) 
I am hoping that we use "Adapters"/"proxies" for FHIR as discussed before. So if the current patient fhir model (provided by openmrs), I should have ability to override that and fetch from a different api (as long as we agree to return a min common agreed model).

internettrans: @gregoryjschmidt Java doesn't work in the browser. Since HAPI is a java library, we can't use it in the browser. Also, HAPI is for integrating with your database, which is not a frontend concern.

The decision about HAPI is one for the backend, not the frontend.

Also, I did research into fhir.js and smart on fhir client-js. They are both well maintained. Their code is good. It has been updated recently. I'm not sure why you are saying fhir.js isn't well maintained, because it seems well maintained to me. I've been impressed with the project in the research and experimentation that I did, including installing the library and trying it out.

internettrans: @angshu without a common model it's impossible to prevent duplicate network requests. I very much agree with you that the common model is troublesome and won't work for many use cases. For the code that doesn't want to use the common model, they can use `authenticatedFetch` or `fhir` to call whichever endpoint they want to however they want to (instead of `getCurrentPatient` which provides only the common model). Doing that causes duplicate network requests, but that is okay sometimes.

text/0013-api-calls-and-fhir.md
```diff
@@ -0,0 +1,76 @@
...
+  adds any needed headers and handles server responses that indicate the user is not logged in.
+- `getCurrentPatient` and other `getCurrent...` functions are provided to share data between disparate microfrontends that
+  need the same data at the same time. They use the frontend route to identify the current patientUuid and then maintain a cache
+  for the shared object. The cache is busted whenever the url changes such that the patientUuid is no longer in the frontend url.
+
+## Reason for decision
+### `authenticatedFetch`
+
+Maintaining an OpenMRS-specific wrapper around `window.fetch()` allows us to make a single code change when
+improving or fixing authentication and other API behavior.
```
fatmali, line 36: @joeldenning  Not sure how this would work out, given that zone.js already has wrapped `window.fetch()`

internettrans: Since authenticated fetch calls window.fetch, zone.js will still be able to make the fetch promise a “Zone Aware promise” and trigger an Angular re-render when the fetch result comes back.

For angular applications, it might be best to inject our own HttpProvider that calls authenticated fetch instead of XHR or window.fetch. See “common practices section” where that is called out.

internettrans:
Merging this because it represents the reality of our code at https://github.com/openmrs/openmrs-esm-api/. We were unable to drive the conversation here on github to get approvals / discussion, so we're just moving forward with it.

PR #14, opened 2019-09-09
Proposal: Configuration (#14)

brandones:


text/0014-configuration.md
```diff
@@ -0,0 +1,78 @@
...
+provide other tooling related to configuration.
+
+
+## Decision
+
+Microfrontend developers should use the official OpenMRS MF Config Library. 
+The Config Library should be a JavaScript module that
+*   Provides an API for modules to define a config schema
+*   Provides an API for modules access config values
+*   Looks for config files in an implementation’s esm-root-config (or using the Import Map)
```
brandones: This is a big open question. Do we expect config files to be specified by esm-root-config or by the import map?

mogoodrich:
I haven't though through the weeds of this, but, in general, making it easy for new microfrontends to config is extremely important, and this proposal seems to be addressing a lot of the pain points in configuration.  We want to support "hot" reconfiguration, but I assume facilitating that becomes much less of an issue in a client-side environment.

Would be interesting in hearing more arguments against parts of this proposal.  (If there are any).

text/0014-configuration.md
fatmali: I prefer root config to be part of esm-root-config mainly because of 1 reason: version control. The import map is currently not versioned so it will be difficult to track the changes overtime to the config files url. The root config is versioned and would be the better alternative.

brandones: The config file wouldn't be contained in the import map, it would be referred to by the import map. It could be managed in version control just like anything else the import map refers to.

text/0014-configuration.md
```diff
@@ -0,0 +1,78 @@
+# Configuration
+- Start Date: 2019/09/09
+- RFC PR:
```
internettrans: This should be a link to this pr:

```suggestion
- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/14
```

text/0014-configuration.md
```diff
@@ -0,0 +1,78 @@
...
+schemas, and loads config files in accordance with that schema. It may also 
+provide other tooling related to configuration.
+
+
+## Decision
+
+Microfrontend developers should use the official OpenMRS MF Config Library. 
+The Config Library should be a JavaScript module that
+*   Provides an API for modules to define a config schema
+*   Provides an API for modules access config values
```
internettrans: grammar nit:

```suggestion
*   Provides an API for modules to access config values
```

text/0014-configuration.md
internettrans: I'm open to either putting the implementation config stuff discussed in this RFC into `@openmrs/esm-root-config` or into a new module called `@openmrs/esm-feature-config` (or similar).

JJ, Brandon, and I talked about it for a while on slack a couple days -- I think I lean towards a new module called `@openmrs/esm-feature-config` so that esm-root-config doesn't continue growing into a catch-all for all important features. The "root-config" refers to the "single-spa root config", as described [in the single-spa docs](https://single-spa.js.org/docs/configuration.html).

One concern brought up regarding the creation of a module for this is that it would require distros to have/maintain one more code repo. I don't view this as a big problem - setting up a distribution/implementation will require maintaining a package.json and a couple repos. And installing dependencies from npm and creating an import map, which essentially will just be an `npm install && npm run build`.

Open to other ideas on this, though.

text/0014-configuration.md
```diff
@@ -0,0 +1,78 @@
...
+*   **Generate config documentation**. It’s hard to keep docs complete and up to date. It would be much easier if nicely-formatted documentation could be generated from the schema.
+*   **Enforce conventions**. The config library can enforce certain conventions about, for example, config namespacing or acceptable ways to reference concepts.
+*   **Validation during config-writing, speed in production**. The config library can execute complex and potentially expensive validation while the implementer is working on the configuration. In production, validation can be turned off to reduce load time.
+*   **Possible future support for other config file types**. While SystemJS provides native support for JSON import, it would be great to allow implementers to write config files that are less verbose, easier to work with, and support comments. Good candidates include HCL and YAML.
+
+For having defaults for all values
```
internettrans, line 46: 👍 👍

text/0014-configuration.md
```diff
@@ -0,0 +1,78 @@
...
+*   Eliminate boilerplate.
+
+For having one combined config file
+*   Since there would be defaults for all values, there will not be config boilerplate to be pulled from external sources. Many microfrontends will probably not need configuration files at all for most implementations. All configurations will have been written by the implementer, and therefore should be fairly short and comprehensible.
+
+For supporting hierarchical configuration
+*   Many organizations have multiple implementations with many similarities and a few differences. They should not be forced to manage config files for each of their implementations separately.
+
+For having the config library handle hierarchies and provide the UI config tool
+*   Configuration hierarchies quickly get messy.
```
internettrans, line 53: agreed

text/0014-configuration.md
```diff
@@ -0,0 +1,78 @@
...
+*   **Generate config documentation**. It’s hard to keep docs complete and up to date. It would be much easier if nicely-formatted documentation could be generated from the schema.
+*   **Enforce conventions**. The config library can enforce certain conventions about, for example, config namespacing or acceptable ways to reference concepts.
+*   **Validation during config-writing, speed in production**. The config library can execute complex and potentially expensive validation while the implementer is working on the configuration. In production, validation can be turned off to reduce load time.
+*   **Possible future support for other config file types**. While SystemJS provides native support for JSON import, it would be great to allow implementers to write config files that are less verbose, easier to work with, and support comments. Good candidates include HCL and YAML.
+
+For having defaults for all values
+*   Eliminate boilerplate.
+
+For having one combined config file
```
angshu: I am not very sure of a single combined config file. Taking ref of Bahmni, where we use app specific configs, I think it can quickly get messy and quite a lot to manage. If each esm provided a means of discovery of such config, then its better that such app configs are version-able, and independently deployable without having to merge onto a single config.

brandones: How about if we can support either one config file or as many config files as the implementer wants to provide, so that the implementer could e.g. start with a single config file and then split it into multiple configs if it becomes unmanagable?

internettrans: The way I've been thinking about it is that it's one ES module git repo but that it could `npm install` many different config files, plus have one or more config files in the source code directly in the repo.

Not sure if this is what you were thinking Brandon or not? I was sort of inferring it from this languaging in the RFC:

>  A **config library** is a code module that allows developers to define config schemas, and loads config files in accordance with that schema.

(note the "loads config file**s**")

This topic is a good one to iron out in the RFC.

brandones: Yes @joeldenning , that makes sense to me.

To be clear, I would expect `npm install`ing config files to pretty much only be used for building off of a organization-specific "base distro," e.g. how the PIH Mexico EMR is built off of PIH EMR.

text/0014-configuration.md
```diff
@@ -0,0 +1,75 @@
...
+provide other tooling related to configuration.
+
+
+## Decision
+
+Microfrontend developers should use the official OpenMRS MF Config Library. 
+The Config Library should be a JavaScript module that
+*   Provides an API for modules to define a config schema
+*   Provides an API for modules to access config values
+*   Provides an API for implementations to provide config files
```
brandones, line 25: Can we just have the config library define an API to "provide" config files, and let implementers decide where to put them and how to reference them? We could establish as "best practice" that the implementation's `esm-root-config` contains

```
// root-config.ts

import ModuleConfig from 'pih-mexico-module-config'
```

Which would provide the config files when executed. This would still allow someone who felt so inclined to provide configs directly in the root config:

```
// root-config.ts

import Config from '@openmrs/esm-config'
import { pihCoreConfig } from `@pih/core-config`
import { mexicoConfig } from './configs/mexico`

Config.provide(pihCoreConfig)
Config.provide(mexicoConfig)
```

internettrans: 👍 yep I think that would be a great way of doing it. Even better would be to maybe provide a name for the config, so that we can give better debugging messages when explaining why a specific config value is the way that it is:

```js
import Config from '@openmrs/esm-config'
import { pihCoreConfig } from `@pih/core-config`
import { mexicoConfig } from './configs/mexico`

// Second argument is name. Alternatively, name could be baked into the config itself which could be better.
Config.provide(pihCoreConfig, 'pihCore')
Config.provide(mexicoConfig, 'mexico)
```

brandones: Oh yeah, I like that a lot.

brandones:
@angshu @fatmali @ojwanganto @joeldenning How we feeling about the "Approve" button?

internettrans (approved):

text/0014-configuration.md
brandones: How can implementers provide config files if they aren't overriding esm-root-config?

Perhaps esm-root-config itself should attempt to import `module-config` from the import map, ignoring if it doesn't find it? And using it as the lowest-priority config file if it does (i.e., if someone both `provide`s a config file in a root config override and in the import map, the values from the former will take priority over the latter)?

internettrans: > How can implementers provide config files if they aren't overriding esm-root-config?

Call `Config.provide()` from any other module.

internettrans: I fear we've gotten confused talking about these things because we haven't shared the same terminology. Might be easier to figure this stuff out as we implement it instead of talking in abstractions in this RFC.

I currently lean towards all the code for all the code for config to be in esm-root-config. And that there is no other module involved, no npm package, nothing. Just code inside of esm-root-config.

brandones: > Call Config.provide() from any other module.

I mean if the implementer isn't writing any code at all. Just throwing together modules and an import map.

> I currently lean towards all the code for all the code for config to be in esm-root-config. And that there is no other module involved, no npm package, nothing. Just code inside of esm-root-config.

Sounds great, I'll do that.

> I fear we've gotten confused talking about these things because we haven't shared the same terminology. Might be easier to figure this stuff out as we implement it instead of talking in abstractions in this RFC.

I agree. Looking forward to sharing the config library once I make some real progress with it.

text/0014-configuration.md
angshu: I am thinking of pluggability, in line with smart-on-fhir apps. Imagine an openmrs app store (e.g. Odoo apps as well) wherefrom I can just add an app, and that app gets downloaded and plugged in. In that case, it would be the case, that the app would bring its own config, which would probably come in their esm module and not requiring any npm build. We should keep our design open. Also reason for also supporting separate config esms for modules.

text/0014-configuration.md
angshu: I like the idea of modules having flexibility of calling config.provide(). 
I don't like the idea of esm-root-config modifications everytime. I prefer each module that provides a configuration to register its configuration, at the time of loadup.

angshu (approved):

fatmali (approved):

brandones:
Thanks, team! @joeldenning , I suppose we can merge whenever (I don't have write access).

internettrans:
Yep! Merged. Also I gave the MF Squad write access to this repo

internettrans:
@brandones you have not accepted the invite to github team called microfrontends squad. This is why you did not have write access. I'm not sure which URL you need to go to in order to accept the invite -- maybe just github.com/openmrs. Or https://github.com/orgs/openmrs/teams. Accepting the invite gives you write access to all the relevant repos, not just this one.

PR #17, opened 2019-10-02
Proposal: Internationalization

fatmali:

### DIFF
--- a/text/0016-internationalization.md
+++ b/text/0016-internationalization.md
@@ -1,0 +1,117 @@
+
+# Internationalization
+
+- Start Date: 2019/09/26
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/17
+	
+
+## Definitions
+
+Localization refers to the adaptation of a product, application or document content to meet the language, cultural and other requirements of a specific target market (a locale).
+
+It is sometimes written as l10n, where 10 is the number of letters between l and n.
+
+  
+
+Internationalization is the design and development of a product, application or document content that enables easy localization for target audiences that vary in culture, region, or language. It is often written i18n, where 18 is the number of letters between i and n.
+
+  
+
+Transifex is a cloud-based localization platform built to help you manage the translation and localization of your app, website, video subtitles, and more. It acts as a repository for your content (think GitHub for translation) and includes tools for developers to get that content into Transifex automatically. Transifex also provides translators a web interface to submit translations.
+
+  
+
+## Decision, including impact on distributions
+
+OpenMRS is implemented throughout the world and is used by speakers of many different languages. As not every user speaks English as a primary language, we will:
+
+1.  Provide an ESM module to translate content in the UI to the user’s specified language.
+    
+2.  The default language is English, this should be made visible and explicit in the code.
+    
+3.  The ESM translation module will be able to translate content **without** connectivity.
+    
+4.  Presently, OpenMRS uses Transifex ([https://www.transifex.com/openmrs/OpenMRS/dashboard/](https://www.transifex.com/openmrs/OpenMRS/dashboard/)) to manage translations. We will continue to use Transifex to support translations for the MF project.
+    
+5.  Possibly find ways to reuse OpenMRS module translations found in the `message.properties` file.
+
+7.  Update mechanism: Once translations are marked as complete, create a mechanism to publish the translated strings to a github repo. 
+    
+
+## Reason for decision
+
+- Most distributions are deployed to areas with no to intermittent connectivity, thus the translations module should be completely offline.
+
+- English is the lingua franca and spoken in most parts of the world.
+
+- Reasons for choosing Transifex:
+	 - Transifex is the most popular internationalization SaaS and offers tools that would make the translations workflow smoother.
+	 - The community is already familiar using transifex for translation services.
+
+
+## Alternatives
+- No translation service at all and default to English for all frontend modules.
+-  Use another service provider built specifically for frontend applications such as [react-i18next](https://github.com/i18next/react-i18next)
+- We could store the translations with each `esm` module instead of having a translations `esm` module.
+- Repo to handle translations service: we could have the translations handled by `openmrs-esm-api`
+
+
+## Common practices (not enforced)
+
+
+### Definitions
+
+Messages - OpenMRS historically used this word to refer to the content that would be replaced.
+
+  
+### Structure of Message File
+
+Here’s an example of how message.properties file looks like:
+
+[https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties](https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties)
+
+
+### Existing API
+
+Here is a [link](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/messagesource/MessageSourceService.java) to the github repo for the interface to the java MessageSourceService in OpenMRS core. It has four methods:
+
+    public String getMessage(String s);
+    
+    public MutableMessageSource getActiveMessageSource();
+    
+    public  void  setActiveMessageSource(MutableMessageSource activeMessageSource);
+    
+    public  void  setMessageSources(Set<MutableMessageSource>  availableMessageSources);
+
+  
+
+### Proposed API
+
+1.  `setPreferredLocale()`: sets the preferred locale of the user, defaults to English if the user has no preferred locale.
+    
+2.  `getMessages(translation)` : takes argument `es-mx` and returns all the messages for that locale.
+    
+3.  `translate(messageVariableName, englishMessage, target-language)` : takes a message and translates it to the target language, if target language is not found, it returns the englishMessage
+    
+
+  
+  
+
+### Example Code
+
+    import { getCurrentUser } from "@openmrs/esm-api";
+    
+      
+    
+    const language = getCurrentUser().getLanguage(); //pseudocode
+    
+    function getMessages(language) {
+    
+    import messages from “@openmrs/esm-intl-messages”
+    
+    }
+    
+      
+    
+    const title = translate(“Patient Search”, es-mx);
+


text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+Here’s an example of how message.properties file looks like:
+
+[https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties](https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties)
+
+
+### Existing API
+
+Here is a [link](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/messagesource/MessageSourceService.java) to the github repo for the interface to the java MessageSourceService in OpenMRS core. It has four methods:
+
+    public String getMessage(String s);
```
internettrans, line 78: easier/better way to get a codeblock in github markdown is to use three ` (backtick) characters in a row. See https://help.github.com/en/articles/creating-and-highlighting-code-blocks. Note that if you do it that way that you can also specify the language after the third back tick in order to get syntax highlighting:

highlighting

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+    
+    public  void  setActiveMessageSource(MutableMessageSource activeMessageSource);
+    
+    public  void  setMessageSources(Set<MutableMessageSource>  availableMessageSources);
+
+  
+
+### Proposed API
+
+1.  `setPreferredLocale()`: sets the preferred locale of the user, defaults to English if the user has no preferred locale.
```
internettrans, line 90: From a slack conversation with @bmamlin a while ago: 

> FYI – the language selection is associated with a user (available languages listed in a global property, user preferred language(s) stored in a user property, and selected language is stored in the session).

Here's a screenshot of the locale being part of the session (`GET /openmrs/ws/rest/v1/session`)
image

A function for setting preferred locale would have to communicate with the backend to update the session. And I don't think we'll have to default to English in the frontend code -- I believe the session *always* returns a locale and so we can just use that one.

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+    
+2.  `getMessages(translation)` : takes argument `es-mx` and returns all the messages for that locale.
```
internettrans, line 92: I don't feel great about this API - getting every single translated string requires that we download every single translated string into the browser. At least some (and maybe many) of the translated strings will likely be lazily downloaded, only as needed by the frontend modules that use them.

I don't think we need this API

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+    
+3.  `translate(messageVariableName, englishMessage, target-language)` : takes a message and translates it to the target language, if target language is not found, it returns the englishMessage
```
internettrans, line 94: Hmm should target-language even be an argument here? Should we just use the locale in the user's session?

If this argument does exist, it should definitely be optional

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+    
+
+  
+  
+
+### Example Code
+
+    import { getCurrentUser } from "@openmrs/esm-api";
```
internettrans, line 102: I think the following code would be simpler and also possible:

```js
import { translate } from '@openmrs/esm-i18n'

translate("transifexProjectIdThing.firstName", "First name")
```

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+### Definitions
+
+Messages - OpenMRS historically used this word to refer to the content that would be replaced.
+
+  
+### Structure of Message File
+
+Here’s an example of how message.properties file looks like:
+
+[https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties](https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties)
```
internettrans, line 71: This file is a java-specific format (see [explanation](https://crunchify.com/java-properties-file-how-to-read-config-properties-values-in-java/)).

I'd prefer us to use a different format that is more friendly to javascript, such as JSON, for the frontend translations. I think it would be okay to parse and use `.properties` files if needed to interop with existing things (could use something like [this npm package](https://www.npmjs.com/package/properties-file) to pull it off), but I don't think it makes sense to put `.properties` files into our frontend code if we don't need to.

internettrans (commented):
👍 nice start on this.

I'd love to see discussion here about some other internationalization issues, including the following:

- standard day format for the locale. (mm/dd/yyyy vs dd/mm/yyyy)
- standard time format (12 hour vs 24 hour format)
- left-to-right vs right-to-left differences in languages. Also top-to-bottom vs bottom-to-top
- currency formatting (`$` vs `£` etc)
- number formatting (`1,000.23` vs `1.000,25`)
- calendar differences (Sunday vs Monday as first day of week, gregorian vs other calendar)
- Any regional timezone related issues
- Differences in physical addresses. How you display them and how you input them.

Additionally here are a few more things to discuss:

1. Should we detect the locale the user has set for the browser and prompt them to update the OpenMRS locale to that locale? See https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages
2. How to address situations like the following. Three separate messageIds??!?!?!? Or just one messageId? If one string, how?
```jsx
<div>
  Some text <span>with some bold</span> in it
</div>
```
3. Impact on design process. How text cannot be assumed to be a certain width or be in English.

text/0016-internationalization.md
jonathandick: I agree with the above

internettrans: Maybe even the following would be best, since this is going to be so ubiquitous:

```js
import { t } from '@openmrs/esm-l10n'

t("messageId", "First name")
```

Another option: in the frontend web community, there's a convention to use a function `__()` for translations (see [example](https://github.com/webpack/webpack/tree/master/examples/i18n)), which we could also adopt.

text/0016-internationalization.md
brandones: Right now the RefApp frontend makes a zillion async API calls for messages, many redundantly. Some of these fail frequently, causing `coreapps.foo.bar.baz` to be all over the page where there should be UI text. I would like to avoid re-creating this problem. Maybe they can be grouped by page or something?

mseaton: Can this be flexible? eg. go to a client side cache/store for these first and then retrieve from server if not found?  Then potentially different consumers could choose to preload the cache or load on demand.

internettrans: Storing in a client side cache could be cool.

I'm thinking that we would make one network request per Transifex *project*. Most definitely not per string.

So there would be an "MF Core" transifex project that would always be available. But then new modules could specify dependencies on their own transifex projects which would be lazily downloaded. Introducing a client-side cache could be interesting, too, to get rid of the need for even those network requests.

text/0016-internationalization.md
dkayiwa: I agree. 👍

text/0016-internationalization.md
internettrans: After thinking about this more, I think that using the browser's built-in cache for network requests would be more than sufficient here. The server for the frontend files can send a `Cache-Control` header and that should be enough to do the trick.

We should be shooting for **a few network requests at most** for translations when a page is loaded. If we get into a scenario where we're making more than 10 network requests just for translations, we've gone wrong.

text/0016-internationalization.md
bmamlin: I'm not sure we can rely on browser locale. While it may be a reasonable default, the use cases we've encountered with OpenMRS are multi-lingual environments and not everyone is using their own device (e.g., users may share a tablet or desktop in a common area). The administrator defines a set of supported languages for the server, users can select from that a subset of their preferred languages, and then we allow them to select their language through the UI. so, I'd expect our translation tooling to use a preferred language from the user's context and, if that is not defined, fall back to the browser's setting.

text/0016-internationalization.md
bmamlin: +1000 to having the default text in code (i.e., using/creating a tool to grep them and create our default English bundle). I'm not a fan of codebases that just put message IDs alone in the code.

text/0016-internationalization.md
internettrans: @bmamlin my point is that the `targetLanguage` argument to the `translate` function is not going to be used because we'll be relying on the OpenMRS locale set in the user profile -- my original comment was questioning the need to allow any string to be translated to any language instead of just always translating to the locale specified in the user profile.

We have decisions to make about how to pull this off technically and allowing for any targetLanguage to be passed in would complicate it. The decisions are what to do at build time vs runtime, and if things are happening at runtime when do we download the translation strings. Note that the translate function here is synchronous, but that downloading extra translations is asynchronous. In order to make that possible, we have to know upfront which strings and language(s) to download before anybody calls the `translate` function.

My vote is for not allowing the targetLanguage to be specified here in the synchronous translate function.

internettrans:
We should decide how to do string interpolation in JSX. This is the main thing that we need to solve in order for this PR and the implementation to go forward. Options are to use [react-i18next's `<Trans>` component](https://react.i18next.com/latest/trans-component), [react-intl's FormattedMessage component](https://github.com/formatjs/react-intl/blob/master/docs/Components.md#formattedmessage), modify/update [react-i18nliner](https://github.com/jenseng/react-i18nliner/) to do what we need, or something else.

internettrans:
I vote we use i18next. Open questions are whether we can get Transifex to output the right format or not, or if we should consider switching to locize.com, which is the service hosted by the i18next people.

jonathandick:
I agree the above from @joeldenning . @burke, others, if locize proves to be an easier/better/same cost option compared to transifex, how strongly do you feel about continuing to use transifex. As I see it, the existing transifex content is not reusable for new projects.

mogoodrich:
@joeldenning is there a particular reason to chose react-i18next over react-intl?  We've been using react-intl in our React Components module.  That being said, if react-i18next is considered better, I have no problem switching.

@jonathandick  what do you mean that "the existing transifex content is not reusable for new projects"?  I'd  hope we could stick with a single localization platform.  If there's a reason that locize.com is significantly better I'm not opposed to moving, just want to make sure it's thought through.  Transifex is currently free for us as a non-profit, and supports outputs in multiple formats.

mogoodrich:
For what it's worth, this seems to be the JSON formats that Transifex supports:



So maybe not what we need.

@jonathandick I responded too quickly, I think I understand your point re: "the existing transifex content is not reusable...".  We definitely couldn't use the existing message.properties files, but if you are translating an English word in a new file that has been previously translated in another translation file, it should suggest that translation for you. So it helps some with consistency.

internettrans:
> is there a particular reason to chose react-i18next over react-intl? We've been using react-intl in our React Components module. That being said, if react-i18next is considered better, I have no problem switching.

i18next works with all frameworks (not just react), react-intl does not. Also, just ergonomically I find react-i18next's `<Trans>` component to be a pretty nice developer experience.

brandones:
Also, shouldn’t be hard at all to transform whatever Transifex will give us into a i18next friendly format. Wrap the tx file update process (tx pull; transform) in an NPM run command.

mogoodrich:
Sounds good @joeldenning ... I haven't been totally in love with react-intl and the fact that it supports multiple frameworks is certainly a big plus.

internettrans:
@fatmali I vote we mark this as ready for review and get it approved. Anything remaining on it?

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+
+Transifex is a cloud-based localization platform built to help you manage the translation and localization of your app, website, video subtitles, and more. It acts as a repository for your content (think GitHub for translation) and includes tools for developers to get that content into Transifex automatically. Transifex also provides translators a web interface to submit translations.
+
+  
+
+## Decision, including impact on distributions
+
+OpenMRS is implemented throughout the world and is used by speakers of many different languages. As not every user speaks English as a primary language, we will:
+
+1.  Provide an ESM module to translate content in the UI to the user’s specified language.
```
brandones, line 28: If I understand correctly, we have translation is not as its own module, but rather integrated into every module. Let's update the text of this document?

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+    
+2.  The default language is English, this should be made visible and explicit in the code.
+    
+3.  The ESM translation module will be able to translate content **without** connectivity.
```
brandones, line 32: Needs clarification. Maybe

```suggestion
3.  It must be possible to serve translated content to clients with only a LAN connection, or without any connection in the case of a standalone deployment.
```

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+    
+4.  Presently, OpenMRS uses Transifex ([https://www.transifex.com/openmrs/OpenMRS/dashboard/](https://www.transifex.com/openmrs/OpenMRS/dashboard/)) to manage translations. We will continue to use Transifex to support translations for the MF project.
+    
+5.  Possibly find ways to reuse OpenMRS module translations found in the `message.properties` file.
+
+7.  Update mechanism: Once translations are marked as complete, create a mechanism to publish the translated strings to a github repo.
```
brandones, line 38:
```suggestion
7.  Translated strings in Transifex should be automatically pulled into the appropriate GitHub repositories.
```

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+4.  Presently, OpenMRS uses Transifex ([https://www.transifex.com/openmrs/OpenMRS/dashboard/](https://www.transifex.com/openmrs/OpenMRS/dashboard/)) to manage translations. We will continue to use Transifex to support translations for the MF project.
+    
+5.  Possibly find ways to reuse OpenMRS module translations found in the `message.properties` file.
+
+7.  Update mechanism: Once translations are marked as complete, create a mechanism to publish the translated strings to a github repo. 
+    
+
+## Reason for decision
+
+- Most distributions are deployed to areas with no to intermittent connectivity, thus the translations module should be completely offline.
```
brandones, line 43:
```suggestion
- Most distributions are deployed to areas with no to intermittent connectivity, thus it should be possible for translations to be available completely offline.
```

text/0016-internationalization.md
```diff
@@ -0,0 +1,117 @@
...
+
+- Reasons for choosing Transifex:
+	 - Transifex is the most popular internationalization SaaS and offers tools that would make the translations workflow smoother.
+	 - The community is already familiar using transifex for translation services.
+
+
+## Alternatives
+- No translation service at all and default to English for all frontend modules.
+-  Use another service provider built specifically for frontend applications such as [react-i18next](https://github.com/i18next/react-i18next)
+- We could store the translations with each `esm` module instead of having a translations `esm` module.
```
brandones, line 55: This is what we are doing, I think?

brandones:
A lot of this RFC would need to be rewritten to adapt it to what we're actually doing. But, actually, we should probably document what we're actually doing on the Wiki -- I think how we handle translation is exactly the sort of cross-cutting, high-level thing that we're wanting to use the Wiki for. So maybe we can rewrite this in a way that can usefully be used in the Wiki. Or we can just close this and write a Wiki page about it.

brandones:
@fatmali Can we close this?

brandones:
Our i18n system is way past proposal stage. I'm closing this.

PR #18, opened 2019-11-12
Proposal: Shared Configuration

brandones:


text/0017-shared-config.md
```diff
@@ -0,0 +1,120 @@
...
+schemas, and loads config files in accordance with that schema. It may also 
+provide other tooling related to configuration.
+
+**Shared config** is configuration values that are shared between distinct modules.
+
+
+## Decision
+
+Config elements that will be shared between multiple modules should be added to a `common`
+config tree. The values provided to this config tree would be made available to any module
```
internettrans, line 24: > The values provided to this config tree would be made available to any module that requested them in its schema.

This means that the `getConfig('@openmrs/esm-login').then(config => {})` would return the common config?

I'm nervous about that. If both the common and module-specific configs have the same key, there will be the issue of merging. I'd prefer this more explicit workflow instead:

```js
const moduleConfig = useConfig()
const commonConfig = useConfig('common')
```

text/0017-shared-config.md
```diff
@@ -0,0 +1,120 @@
...
+A **config library** is a code module that allows developers to define config 
+schemas, and loads config files in accordance with that schema. It may also 
+provide other tooling related to configuration.
+
+**Shared config** is configuration values that are shared between distinct modules.
+
+
+## Decision
+
+Config elements that will be shared between multiple modules should be added to a `common`
```
internettrans, line 23: I'm nervous about this. Common means that there's no "owner" of it - no code and no person who actually understands all of it since it's shared across all codebases and organizations. Seems like an organizational bottleneck to me to try to maintain it.

internettrans (commented):
One alternative - perhaps define the common config schema and maintain it as part of `@openmrs/esm-root-config`? That is the one module that everyone is guaranteed to be using, and is the "root config"

I think it's a natural home for this common config.

text/0017-shared-config.md
brandones: What I’m proposing is that elements of common would need to be explicitly “included” in the config schema. See the example—
`myFoo: common.foo`

text/0017-shared-config.md
brandones: This is a very legitimate concern.

brandones:
> One alternative - perhaps define the common config schema and maintain it as part of `@openmrs/esm-root-config`? That is the one module that everyone is guaranteed to be using, and is the "root config"
> 
> I think it's a natural home for this common config.

Yeah, this is totally reasonable. To be fair, it's pretty similar to what I'm proposing. The questions where we're differing are

1. whether the "common" stuff lives in root-config or module-config (which, like, either way) and
2. whether there's this additional machinery to "include" stuff from common, using the schema, so that clients can do
  ```
  const config = useConfig()
  config.myFoo
  config.bar
  ```
  rather than
  ```
  const rootConfig = useConfig("@openmrs/esm-root-config")
  const config = useConfig()
  rootConfig.foo
  config.bar
  ```

text/0017-shared-config.md
brandones: Added a line below specifying that this would live in `esm-module-config` and be maintained by its maintainer. That wouldn't make it less of a bottleneck though. But I think the bottleneck is inevitable for something intended to be a shared resource like this.

text/0017-shared-config.md
brandones: To clarify: `getConfig('@openmrs/esm-login').then(config => {})` would not return the common config. Given a schema like
```
{
  baz: { default: "bazzz" },
  myFoo: common.foo
}
```
it would return a config like
```
{
  baz: "customBaz",
  myFoo: "customFoo"
}
```

i.e., the common config elements have module-specified insertion points into the module's configuration.

internettrans: When defining the schema, where does `common` come from? It looks like a variable, not a string. Does it need to be imported somewhere?

text/0017-shared-config.md
internettrans: 👍

internettrans:
👍 fair enough. I think putting it into esm-module-config makes sense too. I didn't realize that that was what you were proposing. That sounds good to me. My [question above](https://github.com/openmrs/openmrs-rfc-frontend/pull/18#discussion_r348774374) is the last hold-up I have here before approving this RFC.

text/0017-shared-config.md
brandones: I'm open to recommendations, but I suppose my first inclination would be to make it importable from `@openmrs/esm-module-config`, like
```javascript
import { defineConfigSchema, commonConfig } from "@openmrs/esm-module-config"

defineConfigSchema("my-module", {
  baz: { default: "bazzz" },
  myFoo: commonConfig.foo
})
```

Also very open to suggestions on the name. `common`? `commonConfig`? `commonSchemas`? `commonConfigSchemaComponents`? Not sure what the right level of verbosity is.

Another option might be
```javascript
import { defineConfigSchema } from "@openmrs/esm-module-config"
import { nameConfig } from "@openmrs/esm-module-config/common"

defineConfigSchema("my-module", {
  baz: { default: "bazzz" },
  myFoo: nameConfig
})
```

internettrans: I see. I like the former.

internettrans (approved):
lgtm

brandones:
@fatmali @jonathandick @angshu ?

jonathandick:
LGTM

brandones:
This isn’t likely to be needed in the near term, and this PR is extremely old. I’m going to close this PR. We can re-open if it becomes something we want.

brandones:
This isn’t likely to be needed in the near term, and this PR is extremely old. I’m going to close this PR. We can re-open if it becomes something we want.

PR #19, opened 2019-11-16
RFC: Issue Tracking and Documentation

internettrans:
Looking forward to discussing this -- this is a deviation from how things have been done in the past at OpenMRS.

@bmamlin @dkayiwa @HerbertYiga I would especially love to hear from you on this - I imagine you'll disagree 😄


text/0019-issue-tracking-and-documentation.md
```diff
@@ -0,0 +1,33 @@
...
+## Alternatives
+Up until this RFC was proposed, the OpenMRS community did not use Github issues, readmes, or markdown in any capacity. This discourages collaboration from the open source community and is less accessible to those who are approaching OpenMRS for the first time. When Github issues are turned off, it gives the user of the software the impression that there is no one maintaining it and little hope of making it better. The communication gap between those who maintain the software and those who use the software is widened.
+
+The extremely popular open source project [Babel](https://babeljs.io/blog/) is an example of why using Github is the best way to go to encourage contribution. They tried to push people to a separate issue tracking system for a while but eventually came back to Github.
+
+## Common practices (not enforced)
+- Add CI and other badges to your Github readmes
+- Add installation instructions (`npm install --save @openmrs/esm-login`) to your readme.
+- Add basic usage examples that are fully self contained and actually work.
+- Add API documentation that explains how the user can import the project and use its functions and data.
```
brandones: - Link from the README to relevant wiki pages, and from wiki pages to the README

text/0019-issue-tracking-and-documentation.md
```diff
@@ -0,0 +1,33 @@
+# Issue tracking and documentation
+- Start Date: 2019/11/15
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/119
+
+## Decision, including impact on distributions
+For frontend projects within OpenMRS, the following software tools are used for issue tracking:
+
+- [Github issues](https://help.github.com/en/github/managing-your-work-on-github/about-issues) are enabled and used for general bug reports and questions from the open source community.
```
brandones: Not clear if this includes bug reports within the MF development team? 

What if we treat Issues like a helpdesk / report management solution, and continue to manage all of our tickets and work in JIRA?

Like, should we file bugs to Issues ourselves?

brandones:
I like these ideas. Just needs a bit of clarification, I think.

mseaton:
I think we need to make a distinction between many open source projects that are built for developers (eg. webpack), and those that are built for implementers (eg. OpenMRS).  To that end, I think that documentation aimed at developers, particularly documentation about the codebase itself, belongs in github within the codebase (eg. in the README or associated markdown files).  This allows the code to be self-documenting and for commits to simultaneously change the code and the documentation of it.

However, documentation aimed at implementers belongs in the OpenMRS wiki.  Those users should never need to know where the source of a module lives, but only how to download, install, and use that module (which should be documented in the OpenMRS wiki and other user guides.

As to the use of github issues, my concern would be the fragmentation of change requests between github and JIRA.  I think it could be interesting to open up usage of Github issues if there were clear guidelines around it, but I think this requires a little more discussion.  I'm interested in @bmamlin and @djazayeri input into this.

mogoodrich:
Sounds like a good discussion for Maputo!

internettrans:
> I think that documentation aimed at developers, particularly documentation about the codebase itself, belongs in github within the codebase

> However, documentation aimed at implementers belongs in the OpenMRS wiki

👍 I agree

> As to the use of github issues, my concern would be the fragmentation of change requests between github and JIRA

JIRA is something that the vast majority of the open source world won't know to look at. Additionally, there's no way to report a bug on a specific Github repo in JIRA (and even if there was, it's something the open source community doesn't understand). I think JIRA is much stronger when it's used as a way to coordinate a project within a team across disciplines, but that it's pretty awful for encouraging discussion and contribution by the worldwide open source community.

By turning off Github issues, we are sending the (maybe sometimes accurate?) message that there is no one maintaining the software. I currently maintain >40 open source projects on Github, the majority of which are not within OpenMRS. I also have made more than 3,000 contributions on Github in the last year. In my opinion, if you want contributions from the open source community you definitely need to let people talk to the maintainers of your projects via Github issues. It's like the choice of Git vs Subversion in 2019 -- you go with Git because it's the industry standard.

I'm not saying we'd actually use Github issues for tracking the MF project's progress. I'm just saying that when someone comes across our software that they should be able to contact maintainers via Github issues. Popular open source projects such as Python, Vue, React, Babel, etc all do this.

text/0019-issue-tracking-and-documentation.md
internettrans: I think that it would make sense for GitHub issues to be a way for users to communicate with maintainers. But that maintainers and the teams they are on should continue using JIRA to communicate.

I tried to make that clear in the rfc text, but maybe I should update it to clarify this.

^ this is how large open source projects such as python, vue, and react work. Users talk to maintainers on GitHub, but the maintainers communicate elsewhere when coordinating large projects.

text/0019-issue-tracking-and-documentation.md
internettrans: 👍 added

text/0019-issue-tracking-and-documentation.md
internettrans: I added a bit of languaging here to try to clarify this

brandones: Great, that sounds exactly right to me.

mseaton:
> I'm not saying we'd actually use Github issues for tracking the MF project's progress. I'm just saying that when someone comes across our software that they should be able to contact maintainers via Github issues

This sounds sensible.  It will be interesting to see, if/when we do this, how much it is utilized by those who aren't already connected by the project.  Historically I think we have had a very 1:1 relationship between codebase and JIRA project, but I agree that this has diverged and become harder to enforce and maintain over time, and it can often be hard for someone to know today what project to file an issue under for a particular codebase.  

Would you expect that high level features would be specified and iterated upon in JIRA, and then the actual breakdown into specific technical components would happen in Github issues?  Or would you just expect that Github issues provide a means to communicate with the maintainers of a repository?

internettrans:
> Would you expect that high level features would be specified and iterated upon in JIRA, and then the actual breakdown into specific technical components would happen in Github issues? Or would you just expect that Github issues provide a means to communicate with the maintainers of a repository?

The latter - the most common thing I've seen in large Github projects is to turn on Github issues so people can contact you, but then you use your own tools (JIRA, in this case) for actually coordinating work within the group of maintainers.

brandones (approved):

fatmali (approved):

PR #21, opened 2020-01-21
Proposal: A way to release Openmrs SPAs

nkimaina:

### DIFF
--- a/text/0021-versioning-releases.md
+++ b/text/0021-versioning-releases.md
@@ -11,7 +11,7 @@ OpenMRS SPA: refers to an OpenMRS Micro-Frontend app. An example is openmrs-esm-
 
 CDN: Content Distribution Network such as jsdelivr. See https://www.jsdelivr.com/
 
-MAJOR Version: version when you make incompatible API changes
+MAJOR Version: version where there are incompatible API or configuration API changes, or major UI changes that are enabled by default
 
 MINOR Version: version when you add functionality in a backwards compatible manner
 
@@ -24,9 +24,8 @@ The releases should have a maintenance branch for bug fixes and patches so that 
 
 I propose that we have maintenance release branches for both Major and Minor releases, as the motivation for this is to allow implementors/distributions control on when to start using a certain feature, that could be in a certain minor or major release. 
 
-At this time, OpenMRS MF Squad will not maintain older released versions in terms of porting back features and bug fixes to those releases. An implementor/distribution who wishes to benefit from bug fixes, and is on an earlier major or minor release will have to back-port those bug fixes to the older release themselves. Such an individual will need write access to the Github repo hosting the SPA application of interest. 
-
-When to do a major or minor release will be request-driven. An implementor/distribution who may wish to benefit from features not yet released will need to request the Micro-Frontends Squad for a certain feature to be released, in a minor or major version as seen fit.  
+The OpenMRS Microfrontend Squad will not maintain release branches. Implementers or organizations who wish to maintain release branches may do so. They 
+should be granted write access to the GitHub repo hosting the ESM of interest.
 
 Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS Community, so that implementors have stable resources to choose from. 
 
@@ -42,4 +41,3 @@ Tagged releases should be distributed via NPM as well, to support development en
 Each implementor/distribution to have a fork of the OpenMRS SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distributions to diverge, weakening the OpenMRS Community in the process. 
 
 ## Common practices (not enforced)
-- Consider adapting semantic versioning(SemVer), as SemVer allows implementors to more easily assess the work involved in upgrading an application. See https://semver.org/
--- a/text/0021-versioning-releases.md
+++ b/text/0021-versioning-releases.md
@@ -1,0 +1,26 @@
+# Releasing and Versioning of Openmrs SingleSpa Apps
+- Start Date: 2020/01/21
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
+
+## Decision
+Openmrs spa applications should have releases that are versioned so that each implementor can have control on when to upgrade to the latest changes. 
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
+
+Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS, as is currently the case with the openmrs single-spa test environment hosted in digital ocean spaces, so that implementors get only stable resources to choose from. 
+
+We should adapt semantic versioning, as it is the most widely accepted versioning scheme for releases. See https://semver.org/ 
+
+## Definition
+
+## Reason for decision
+- Implementors and various distributions will want to control when features are added to their distributions. 
+- In addition to the reason 1) above, implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their versions of Openmrs SPAs, through the use of maintenance branches for releasing.
+- Distribution via CDN is desirable as it is most mainstream way for distributing web-app assets.    
+
+## Alternatives
+Each implementor/distribution to have a folk of the Openmrs SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distros to diverge, Openmrs not benefiting from community support in the process. 
+
+## Common practices (not enforced)
+- Have maintenance branches for open-source projects
+- Have LTS in terms of years defined for this releases. 
--- a/text/0021-versioning-releases.md
+++ b/text/0021-versioning-releases.md
@@ -2,25 +2,44 @@
 - Start Date: 2020/01/21
 - RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
 
+## Definition
+OpenMRS: refers to the openmrs community.
+
+Micro-Frontends Squad: refers to the OpenMRS Community members responsible for creating and maintaining OpenMRS SPAs, under the OpenMRS Micro-Frontends Project.
+
+OpenMRS SPA: refers to an OpenMRS Micro-Frontend app. An example is openmrs-esm-patient-chart. See https://github.com/openmrs/openmrs-esm-patient-chart
+
+CDN: Content Distribution Network such as jsdelivr. See https://www.jsdelivr.com/
+
+MAJOR Version: version when you make incompatible API changes
+
+MINOR Version: version when you add functionality in a backwards compatible manner
+
+PATCH Version: version when you make backwards compatible bug fixes
+
 ## Decision
-Openmrs spa applications should have releases that are versioned so that each implementor can have control on when to upgrade to the latest changes. 
+OpenMRS SPA applications should have releases that are versioned so that each implementor/distribution can decide when to upgrade to a certain version, containing features that the implementor/distribution needs. 
 
 The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
 
-Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS, as is currently the case with the openmrs single-spa test environment hosted in digital ocean spaces, so that implementors get only stable resources to choose from. 
+I propose that we have maintenance release branches for both Major and Minor releases, as the motivation for this is to allow implementors/distributions control on when to start using a certain feature, that could be in a certain minor or major release. 
 
-We should adapt semantic versioning, as it is the most widely accepted versioning scheme for releases. See https://semver.org/ 
+At this time, OpenMRS MF Squad will not maintain older released versions in terms of porting back features and bug fixes to those releases. An implementor/distribution who wishes to benefit from bug fixes, and is on an earlier major or minor release will have to back-port those bug fixes to the older release themselves. Such an individual will need write access to the Github repo hosting the SPA application of interest. 
 
-## Definition
+When to do a major or minor release will be request-driven. An implementor/distribution who may wish to benefit from features not yet released will need to request the Micro-Frontends Squad for a certain feature to be released, in a minor or major version as seen fit.  
+
+Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS Community, so that implementors have stable resources to choose from. 
+
+Tagged releases should be distributed via NPM as well, to support development environments and implementations relying on Packmap tool (see https://github.com/openmrs/packmap) for deployment in an offline site.
 
 ## Reason for decision
-- Implementors and various distributions will want to control when features are added to their distributions. 
-- In addition to the reason 1) above, implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their versions of Openmrs SPAs, through the use of maintenance branches for releasing.
-- Distribution via CDN is desirable as it is most mainstream way for distributing web-app assets.    
+- OpenMRS implementors and various distributions will want to decide on when features are added to their various distributions, achieved through versioning of releases. 
+- Implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their major/minor versions of OpenMRS SPAs. This is done through patch releases, that are only possible through use of release branching. 
+- For implementors relying on the internet to access their OpenMRS application, distribution via CDN is desirable, as it brings web-app assets for  application closer to the client accessing it.
+- Distributions via NPM is desirable as it allows developers to benefit from 'typings' when coding, and for the Packmap tool to be able to create the Openmrs SPA Module artifact for use with offline sites.    
 
 ## Alternatives
-Each implementor/distribution to have a folk of the Openmrs SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distros to diverge, Openmrs not benefiting from community support in the process. 
+Each implementor/distribution to have a fork of the OpenMRS SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distributions to diverge, weakening the OpenMRS Community in the process. 
 
 ## Common practices (not enforced)
-- Have maintenance branches for open-source projects
-- Have LTS in terms of years defined for this releases. 
+- Consider adapting semantic versioning(SemVer), as SemVer allows implementors to more easily assess the work involved in upgrading an application. See https://semver.org/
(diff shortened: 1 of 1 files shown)


text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Definition
+
+## Reason for decision
+- Implementors and various distributions will want to control when features are added to their distributions. 
+- In addition to the reason 1) above, implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their versions of Openmrs SPAs, through the use of maintenance branches for releasing.
+- Distribution via CDN is desirable as it is most mainstream way for distributing web-app assets.    
+
+## Alternatives
+Each implementor/distribution to have a folk of the Openmrs SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distros to diverge, Openmrs not benefiting from community support in the process.
```
ibacher: "fork" rather than "folk". I'm not clear on what the "for the purposes of releasing ONLY" clause is adding. "to diverge, Openmrs..." should probably read "to diverge from Openmrs, not benefiting...".

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
...
+## Reason for decision
+- Implementors and various distributions will want to control when features are added to their distributions. 
+- In addition to the reason 1) above, implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their versions of Openmrs SPAs, through the use of maintenance branches for releasing.
+- Distribution via CDN is desirable as it is most mainstream way for distributing web-app assets.    
+
+## Alternatives
+Each implementor/distribution to have a folk of the Openmrs SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distros to diverge, Openmrs not benefiting from community support in the process. 
+
+## Common practices (not enforced)
+- Have maintenance branches for open-source projects
```
ibacher: "open-source projects" -> "SPA applications"

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
...
+- Have LTS in terms of years defined for this releases.
```
ibacher: "Have a defined long-term support window for releases"

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
...
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
+
+## Decision
+Openmrs spa applications should have releases that are versioned so that each implementor can have control on when to upgrade to the latest changes. 
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
+
+Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS, as is currently the case with the openmrs single-spa test environment hosted in digital ocean spaces, so that implementors get only stable resources to choose from. 
+
+We should adapt semantic versioning, as it is the most widely accepted versioning scheme for releases. See https://semver.org/
```
ibacher: We should probably have a stronger reason for landing on SemVer than just "it is the most widely accepted versioning scheme". Among other things, this decision is both central to the point of this RFC and a break with how the OpenMRS community currently manages versions, though arguably OpenMRS as a whole should move towards adopting SemVer.

I would favour something like: "as SemVer allows implementors to more easily assess the work involved in upgrading an application".

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
+# Releasing and Versioning of Openmrs SingleSpa Apps
+- Start Date: 2020/01/21
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
+
+## Decision
+Openmrs spa applications should have releases that are versioned so that each implementor can have control on when to upgrade to the latest changes. 
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
+
+Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS, as is currently the case with the openmrs single-spa test environment hosted in digital ocean spaces, so that implementors get only stable resources to choose from.
```
ibacher: "digital ocean" -> "Digital Ocean". Do we want an RFC that refers to what the OpenMRS SPA test environment is "currently" doing or even to this environment at all? I think it's sufficient to say "Releases should be tagged and the artifacts distributed via a CDN maintained by OpenMRS, so that implementors have stable resources to choose from".

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
+# Releasing and Versioning of Openmrs SingleSpa Apps
+- Start Date: 2020/01/21
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
+
+## Decision
+Openmrs spa applications should have releases that are versioned so that each implementor can have control on when to upgrade to the latest changes.
```
ibacher: "SPA applications" is redundant. "can have control on" -> "has control over". "when" -> "when and whether". "the latest changes" -> "more recent versions".

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
+# Releasing and Versioning of Openmrs SingleSpa Apps
+- Start Date: 2020/01/21
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/1
+
+## Decision
+Openmrs spa applications should have releases that are versioned so that each implementor can have control on when to upgrade to the latest changes. 
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes.
```
ibacher, line 23: At which level should these maintenance branches be maintained? Should a SPA maintain a branch for every minor version or only for major versions?

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,26 @@
...
+
+Releases should be tagged, and the artifacts distributed via a CDN maintained by OpenMRS, as is currently the case with the openmrs single-spa test environment hosted in digital ocean spaces, so that implementors get only stable resources to choose from. 
+
+We should adapt semantic versioning, as it is the most widely accepted versioning scheme for releases. See https://semver.org/ 
+
+## Definition
+
+## Reason for decision
+- Implementors and various distributions will want to control when features are added to their distributions. 
+- In addition to the reason 1) above, implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their versions of Openmrs SPAs, through the use of maintenance branches for releasing.
```
ibacher: "In addition to the reason 1) above" - remove this text

text/0021-versioning-releases.md
ibacher: Should this RFC make a decision on how bugfixes are addressed vis-a-vis more recent versions? I.e., should we favour bugfixes against 2.x which are backported to 1.x (this is something like the current OpenMRS model) or bugfixes against 1.x which are "forward-ported" to 2.x (this involves more work for the maintainers, but makes it easier for implementors to provide bugfixes).

ibacher (commented):
1. This document uses different capitalisations of both OpenMRS and SPA. These should be unified. I'd favour "OpenMRS" and "SPA".

1. "OpenMRS" is used as a term to variously refer to (a) the OpenMRS community and (b) whatever we are calling the version of the OpenMRS platform + the Single SPA module (is there an accepted name for this?). We should clearly distinguish these two things.

text/0021-versioning-releases.md
mseaton: @ibacher OpenMRS has standardized on SemVer I believe, it just might not be fully enforced or applied retroactively to everything

ibacher: @mseaton Fair. I still think the RFC can provide a stronger rationale for this decision.

text/0021-versioning-releases.md
brandones: If we are going to do release branching, it should only be for major versions.

brandones:
This is about versioning ESMs? They are already versioned according to semvar.

We do not have release branches, and I do not think that we should take on the responsibility of backporting bug fixes, which will reduce our velocity, until there are multiple organizations using MFE in production, and at least one of them decides that they want to stay on an old version of some ESM. That is the point at which we should be considering this.

text/0021-versioning-releases.md
nkimaina: I think we should have release branches for both, as the whole point is to allow implementors to decide when to upgrade to see new features, as well as allow them to benefit from bug-fixes.

nkimaina: The alternative will be to do feature flags, which I think will get messy for an open-source project.

brandones: The kinds of features that people would generally want to disable should be toggled in the config.

For everything else, organizations should ask themselves -- "are we really that special?" And if so, they should introduce a new config parameter.

Backporting bug fixes to every minor version:
- is not justifiable based on current adoption patterns
- will be a huge pain for developers
- will discourage us from releasing, knowing that every release is a new maintenance nightmare we will have to live with until it is EOL'd
- will reduce forward velocity on MFE before it has even had the opportunity to prove itself, at a time when organizations still consider dedicating devs to MFE a risky proposition

For major versions, there is much more of an argument to be made for release branching, since the definition of a major version is one that introduces breaking changes -- i.e., the upgrade requires some effort, and organizations often take a while to make that happen.

nkimaina: If it was library that we are building, then what you are saying makes perfect sense. These front-end features are client-facing, users interact with them, there's training that goes with it, and decisions about workflows that are made by various stakeholders before a feature is introduced. We need to allow an organisation to benefit from bug-fixes and patches, before they are ready to let their users use a new feature or a modification of an existing feature. Effort to back-port bugs to certain release-branches could be left to such an organisation. Configs to support 'timing of release of a certain feature to a certain group of users' will require coordination each release cycle, and effort to maintain anyway, and can get ugly for an open-source project. 
All this said, I think it would be safer to allow creation of these minor release branches on demand, and leaving the responsibility of maintaining it to an organisation that needs it badly. Restricting it will only promote organisations to create forks when they get cornered with a bug that affects them.

nkimaina: We still need the configs that are there by design, to allow an application to be configured to an organisation's  needs, but requiring that every feature to have these configs for the purposes of release adds complexities to the entire application.

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
+## Definition
+OpenMRS: refers to the openmrs community.
+
+Micro-Frontends Squad: refers to the OpenMRS Community members responsible for creating and maintaining OpenMRS SPAs, under the OpenMRS Micro-Frontends Project.
+
+OpenMRS SPA: refers to an OpenMRS Micro-Frontend app. An example is openmrs-esm-patient-chart. See https://github.com/openmrs/openmrs-esm-patient-chart
+
+CDN: Content Distribution Network such as jsdelivr. See https://www.jsdelivr.com/
+
+MAJOR Version: version when you make incompatible API changes
```
brandones: Since most MFEs don't have an API, I'm going to suggest

```suggestion
MAJOR Version: version where there are incompatible API or configuration API changes, or major UI changes that are enabled by default
```

text/0021-versioning-releases.md
brandones: Really good points, @nkimaina . I'm in agreement with letting organizations create and maintain release branches according to their needs.

I'm also going to propose some changes to our definition of major and minor versions, which might help accommodate the needs you are designing for.

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
+
+MINOR Version: version when you add functionality in a backwards compatible manner
+
+PATCH Version: version when you make backwards compatible bug fixes
+
+## Decision
+OpenMRS SPA applications should have releases that are versioned so that each implementor/distribution can decide when to upgrade to a certain version, containing features that the implementor/distribution needs.
```
brandones, line 21: I'd delete this line -- SPA Applications already are versioned according to SemVar.

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
+PATCH Version: version when you make backwards compatible bug fixes
+
+## Decision
+OpenMRS SPA applications should have releases that are versioned so that each implementor/distribution can decide when to upgrade to a certain version, containing features that the implementor/distribution needs. 
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
+
+I propose that we have maintenance release branches for both Major and Minor releases, as the motivation for this is to allow implementors/distributions control on when to start using a certain feature, that could be in a certain minor or major release. 
+
+At this time, OpenMRS MF Squad will not maintain older released versions in terms of porting back features and bug fixes to those releases. An implementor/distribution who wishes to benefit from bug fixes, and is on an earlier major or minor release will have to back-port those bug fixes to the older release themselves. Such an individual will need write access to the Github repo hosting the SPA application of interest.
```
brandones:
```suggestion
The OpenMRS Microfrontend Squad will not maintain release branches. Implementers 
or organizations who wish to maintain release branches may do so. They 
should be granted write access to the GitHub repo hosting the ESM of interest.
```

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
+
+I propose that we have maintenance release branches for both Major and Minor releases, as the motivation for this is to allow implementors/distributions control on when to start using a certain feature, that could be in a certain minor or major release. 
+
+At this time, OpenMRS MF Squad will not maintain older released versions in terms of porting back features and bug fixes to those releases. An implementor/distribution who wishes to benefit from bug fixes, and is on an earlier major or minor release will have to back-port those bug fixes to the older release themselves. Such an individual will need write access to the Github repo hosting the SPA application of interest. 
+
+When to do a major or minor release will be request-driven. An implementor/distribution who may wish to benefit from features not yet released will need to request the Micro-Frontends Squad for a certain feature to be released, in a minor or major version as seen fit.
```
brandones: I think we should feel free to version and release as often as our progress warrants. If we're getting a lot of requests to "please release X," that will mean we are not releasing often enough. I would just get rid of this line, since what I'm suggesting is what we already do.

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
+- Implementors would like to contribute bug fixes and benefit from the bug fixes without necessary upgrading their major/minor versions of OpenMRS SPAs. This is done through patch releases, that are only possible through use of release branching. 
+- For implementors relying on the internet to access their OpenMRS application, distribution via CDN is desirable, as it brings web-app assets for  application closer to the client accessing it.
+- Distributions via NPM is desirable as it allows developers to benefit from 'typings' when coding, and for the Packmap tool to be able to create the Openmrs SPA Module artifact for use with offline sites.    
+
+## Alternatives
+Each implementor/distribution to have a fork of the OpenMRS SPAs that they are using, for the purposes of releasing ONLY. This is a less desirable approach as it encourage implementors/contributors/distributions to diverge, weakening the OpenMRS Community in the process. 
+
+## Common practices (not enforced)
+- Consider adapting semantic versioning(SemVer), as SemVer allows implementors to more easily assess the work involved in upgrading an application. See https://semver.org/
```
brandones: We do use semvar.

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,43 @@
...
+MINOR Version: version when you add functionality in a backwards compatible manner
+
+PATCH Version: version when you make backwards compatible bug fixes
+
+## Decision
+OpenMRS SPA applications should have releases that are versioned so that each implementor/distribution can decide when to upgrade to a certain version, containing features that the implementor/distribution needs. 
+
+The releases should have a maintenance branch for bug fixes and patches so that implementors running a particular version of an app can benefit from recent bug fixes. 
+
+I propose that we have maintenance release branches for both Major and Minor releases, as the motivation for this is to allow implementors/distributions control on when to start using a certain feature, that could be in a certain minor or major release.
```
brandones: Can we be specific about when release branches are created? Can we say that they should be created from the release-tagged commit by whoever decides they are going to maintain that branch? Then we’ll avoid the clutter of a million release branches no one cares about.

brandones:
This RFC is approaching one month old. I feel like it's awfully close. @openmrs/microfrontends-squad can we get some more reviews/approvals on this?

text/0021-versioning-releases.md
nkimaina: Sorry didn't see the comment on time. Do you mean something like 'An implementor or distribution that requires a certain maintenance branch, either for a minor or major release is free to create it, by branching from the release-tagged commit' ?

nkimaina:
@ojwanganto Have you seen this? Will the proposal work for KenyaEMR?

text/0021-versioning-releases.md
brandones: Yeah, that's perfect!

brandones (approved):

ibacher (approved):

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
```
bmamlin: I'd suggest avoiding first person like this to simplify adoption of the RFC – i.e., write an RFC from that perspective that it has already been adopted & merged.

text/0021-versioning-releases.md
```diff
@@ -0,0 +1,45 @@
...
+
+An implementor or distribution that requires a certain maintenance branch, either for a minor or major release is free to create it, by branching from the release-tagged commit.
+
+The OpenMRS Microfrontend Squad will not maintain release branches. Implementers or organizations who wish to maintain release branches may do so. They 
+should be granted write access to the GitHub repo hosting the ESM of interest.
```
bmamlin, line 32: It might be worth mentioning that development should always occur on the master branch with selected changes _within master_ backported to release branches as needed. Making changes directly to a release branch and forwardporting those changes to master is discouraged and should only happen – if ever – in extreme circumstances.

dkayiwa: I completely agree!

text/0021-versioning-releases.md
nkimaina: Agreed.

text/0021-versioning-releases.md
nkimaina: Agreed

bmamlin (approved):

brandones:
:tada:

PR #22, opened 2020-02-07
Proposal: Version definitions

brandones:


bmamlin:
Do we need RFCs for existing OpenMRS conventions? If so, I think this RFC could be simplified to:

> OpenMRS frontend ESMs should be versioned according to the [OpenMRS Versioning conventions](https://wiki.openmrs.org/x/BA8z).

brandones:
This proposal differs from SemVar in two key respects:
- The notion that the config schema is an API (which versioning should respect) is made explicit
- Major UI changes are added as a condition for major versioning

It additionally differs from the OpenMRS conventions in that
- There is no `.database` nor `[-modifier]` part of the version (though I think people should be welcome to use modifiers like "alpha" as they see fit)

Perhaps I should trim down the RFC to really highlight those differences, and just refer to SemVar for the rest?

bmamlin:
> Perhaps I should trim down the RFC to really highlight those differences, and just refer to SemVar for the rest?

Agreed. While I only represent one data point, I saw the RFC as just a duplicate of SemVar or the OpenMRS Versioning convention and your response above was both concise & very informative for me. Maybe you could set a record for the smallest RFC. 😃

brandones:
Well, I fixed it up, but I did not succeed in making it shorter. But reading the OpenMRS Versioning conventions made me realize that we should also have versioning respect backend dependencies.

bmamlin (approved):

text/0022-versions.md
```diff
@@ -0,0 +1,60 @@
...
+provide guidelines for versioning OpenMRS Modules. However, they do not
+usefully apply to ESMs.
+
+## Reason for decision
+
+OpenMRS frontend ESMs do not, for the most part, center their APIs. Rather,
+they are built with a focus on UI features. For most of them, the
+[Configuration Schema](https://github.com/openmrs/openmrs-rfc-frontend/blob/master/text/0014-configuration.md)
+is the most-used API. Due to these considerations, some extension of
+SemVar is warranted, in order to provide adequate guidance for versioning
```
ibacher: SemVer

text/0022-versions.md
```diff
@@ -0,0 +1,60 @@
...
+  default different from the previous UI
+- A backend module dependency has been introduced or upgraded across a minor version
+
+The key ideas here are that
+
+- The configuration schema is an API, which versioning should respect
+- Major UI changes are added as a condition for major versioning
+- ESM versions should be sensitive to backend module requirements
+
+Apart from the differences made explicit in this RFC, SemVar should be followed.
```
ibacher: SemVer

text/0022-versions.md
```diff
@@ -0,0 +1,60 @@
...
+frontend ESMs.
+
+Additionally, implementers should expect to be able to upgrade ESMs across minor
+versions without having to worry about backend module dependencies changing.
+
+## Alternatives
+
+- Version strictly according to the public APIs (and potentially the configuration schemas)
+    of ESMs, but not backend module dependencies or UI changes. This would be more true
+    to the text of SemVar.
```
ibacher: SemVer

text/0022-versions.md
```diff
@@ -0,0 +1,60 @@
...
+OpenMRS frontend ESMs should be versioned according to
+[Semantic Versioning](https://semver.org/), with the following modification
+to the definition of a major version:
+
+A **Major** version is one in which **any** of the following have taken place
+- Backward-incompatible changes have been made to the API, if the ESM exposes an API
+- Backward-incompatible changes have been made to the Configuration API,
+  such that upgrading might cause a previously valid configuration to be invalid
+- A functional and substantive UI change has been introduced, which is by
+  default different from the previous UI
```
ibacher: I think this would be better phrased as "A functional and substantive UI change has been introduced such that the UI differs from the previous UI using the default settings"

text/0022-versions.md
```diff
@@ -0,0 +1,60 @@
...
+    to the text of SemVar.
+- Invent some other versioning rationale.
+- Abolish versions. Nothing but master and its irregular branches and forks. Has
+    the potential to make life solitary, poor, nasty, brutish, and short.
+
+## Common practices (not enforced)
+
+- When introducing a major new feature or UI change that is toggled in configuration, consider
+  releasing first a minor version that has the feature disabled by default,
+  and then enabling it by default in the next major version.
```
ibacher: "consider first releasing"

ibacher (approved):
I like this RFC. Particularly the bit about version anarchy. I've suggested some minor wording improvements, but feel free to ignore them.

text/0022-versions.md
brandones:
```suggestion
SemVer is warranted, in order to provide adequate guidance for versioning
```

text/0022-versions.md
brandones: Yes, that is better, thank you!

brandones: Sorry, I changed my mind and decided it wasn't better. Though you're right about the grammatical problem in the referent of "which." I changed it a bit.

text/0022-versions.md
```diff
@@ -0,0 +1,60 @@
...
+to the definition of a major version:
+
+A **Major** version is one in which **any** of the following have taken place
+- Backward-incompatible changes have been made to the API, if the ESM exposes an API
+- Backward-incompatible changes have been made to the Configuration API,
+  such that upgrading might cause a previously valid configuration to be invalid
+- A functional and substantive UI change has been introduced and that change is
+  present in the default configuration
+- A backend module dependency has been introduced or upgraded across a minor version
+
```
jonathandick, line 17: Can we also include:

> A frontend module dependency has been introduced or upgraded across a minor version

jonathandick (changes requested):

text/0022-versions.md
brandones: Can you provide an example of when that would happen?

jonathandick: Nevermind, handled by semver.

jonathandick (approved):

PR #23, opened 2020-04-17
First draft of 0023-readme.md

brandones:


text/0023-readme.md
```diff
@@ -0,0 +1,68 @@
...
+- Start Date: 2020/04/17
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/23
+
+## Decision, including impact on distributions
+OpenMRS ESMs will maintain documentation of their features, APIs,
+configuration schema, and compatibility concerns in their README.
+
+The OpenMRS Wiki page
+[List of frontend modules](https://wiki.openmrs.org/pages/viewpage.action?pageId=224527568)
+will link directly to the README in the ESM's GitHub repo's page. The
```
FlorianRappl: Can we make it "will link to the GitHub repository of the ESM"?

FlorianRappl (approved):
LGTM!

nkimaina (approved):

text/0023-readme.md
brandones: Yep! Thanks for reviewing!

ibacher:
Some questions / suggestions / comments:

* Would it be worth borrowing the current structure from the Wiki to cover the brief description? I'm thinking in particular of the "What is this?" and "How do I use it?" sections. It might be useful to have such prompts to encourage a consistent documentation structure.
* Has any consideration been made of the use of GitHub pages?
* Should this RFC or a subsequent one make non-normative suggestions as to the formatting of API docs?
* Might it make sense to include a non-normative exemplar of documentation for an ESM?

brandones:
>   Would it be worth borrowing the current structure from the Wiki to cover the brief description? I'm thinking in particular of the "What is this?" and "How do I use it?" sections. It might be useful to have such prompts to encourage a consistent documentation structure.

Sure! I don't want to be too prescriptive about the README structure here — what if we add that as a "suggested practice" at the end, and then establish the precedent in all of the existing READMEs?

>   Has any consideration been made of the use of GitHub pages?

Nope! Hadn't crossed my mind. What would we use it for?

>   Should this RFC or a subsequent one make non-normative suggestions as to the formatting of API docs?

I think we have a lot of work to do figuring out how to write any API documentation at all, and figuring out if we can get automated API documentation generation working. But also I guess I'm inclined to let developers figure out how to do API documentation best for their own situation (while having a normative way available that's clean and easy and works well 80% of the time).

>   Might it make sense to include a non-normative exemplar of documentation for an ESM?

I think part of this RFC will be adapting the documentation to comply with it -- moving out of the wiki, documenting APIs, etc. If we don't completely screw it up, all of the core ESMs should serve that purpose nicely.

ibacher:
@brandones Thanks for your thoughtful replies!

> What if we add that as a "suggested practice" at the end, and then establish the precedent in all of the existing READMEs?

I think that actually addresses basically all of my comments. Yes! This!

> Nope! Hadn't crossed my mind. What would we use it for?

It may not apply. I hadn't really looked to closely at the scope for most of the ESMs (patient chart, though, looks to be quite complex). I was just thinking it's a middle-ground between the README and a full-fledged documentation site like ReadTheDocs if an ESM needed a somewhat longer technical description than comfortably fit in a README.

brandones:
> It may not apply. I hadn't really looked to closely at the scope for most of the ESMs (patient chart, though, looks to be quite complex). I was just thinking it's a middle-ground between the README and a full-fledged documentation site like ReadTheDocs if an ESM needed a somewhat longer technical description than comfortably fit in a README.

Nice, that makes sense. I think we can leave that for maintainers to figure out as they see fit.

drizzentic (approved):

drizzentic (approved):

brandones:
Oh my god, I just realized that the part of this I care about is already covered by https://github.com/openmrs/openmrs-rfc-frontend/blob/master/text/0019-issue-tracking-and-documentation.md . We just never carried out the migration.

> For frontend projects within OpenMRS, the following software tools are used for documentation:
> - Github readmes with markdown files are used to document installation, usage, and API documentation for projects.
> - OpenMRS wiki is used for in-depth explanations of concepts, tutorials, and examples.

I'm going to close this. But let's migrate!

brandones:
Maybe that was brash; there is a good deal of detail that's part of this that's not part of #19 . We need one more approval from the squad (or dissenters, speak up!). @ibacher @bmamlin @donaldkibet @ivange94  @dkayiwa @ojwanganto

donaldkibet (approved):

PR #24, opened 2020-04-27
First draft of 0024-setup.md

FlorianRappl:
Thoughts and ideas for simplifying and improving our dev setup per repository used for the microfrontends.

Up for discussion :rocket:!

### DIFF
--- a/text/0024-setup.md
+++ b/text/0024-setup.md
@@ -1,0 +1,87 @@
+# Project title
+- Start Date: 2020/04/23
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/24
+
+## Decision, including impact on distributions
+All frontend modules should use a common visual coding style to simplify
+collaboration and streamline discussions.
+
+Consequently, repositories in the `openmrs-esm-*` space should opt-in to use a
+set of common files:
+
+- `.editorconfig`
+- `prettier.config.js`
+- `.eslintrc.json`
+
+Presumably, the following contents make sense.
+
+For the `.editorconfig` file we propose to use:
+
+```plain
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+max_line_length = null
+```
+
+For the `prettier.config.js` file we propose to use:
+
+```js
+module.exports = {
+  printWidth: 120,
+  singleQuote: true,
+  trailingComma: 'all',
+  bracketSpacing: true,
+  parser: 'typescript',
+  semi: true,
+  jsxBracketSameLine: true,
+};
+```
+
+For the `.eslintrc.json` file we propose to use:
+
+```json
+{
+  "extends": [
+    "ts-react-important-stuff",
+    "plugin:prettier/recommended"
+  ],
+  "parser": "babel-eslint"
+}
+```
+
+## Definition
+A **visual coding style** defines how the syntax of a programming language is
+expressed in terms of flexible layers, mostly optional parentheses, white space,
+and line breaks.
+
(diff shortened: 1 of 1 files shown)


brandones:
All of our repositories already do use prettier and eslint. The browserconfig is a nice idea. What's the motivation for this?

FlorianRappl:
> What's the motivation for this?

All repositories I've checked out do not contain any `.editorconfig` or `prettier.config.js` file. Only the `openmrs-esm-patient-chart-widgets` contained a `.prettierrc` file - but it was empty.

Coming up with a common config makes it easier for the respective plugins (e.g., VS Code Prettier Formatter) to do the right thing. Otherwise it potentially falls back to custom formatting.

brandones:
Yeah, we've just been using prettier out of the box. I think it does fine, but I'm not opposed to configuring it. But I think it's more a matter of taste and not something I'm inclined to make a community-wide mandate about.

Sorry, when I said "browserconfig" above, I meant "editorconfig." I just tell VSCode to use prettier for formatting. Is there an advantage to providing an editorconfig when already using prettier?

FlorianRappl:
> Is there an advantage to providing an editorconfig when already using prettier?

The both work side-by-side and are not exclusive. Actually, prettier picks up the rules from `.editorconfig`. The latter defines actually how source files (independent of MD, JS, ... etc. - even though it could be fine-grained) look like (e.g., 2 spaces, finished blank line).

So while prettier is there for actually syntax formatting rules, editorconfig is there for general text file formatting. As such if I would have to pick one, I would always go for editorconfig.

FlorianRappl:
**One remark - just to be clear**:
I have no shares in the particular rules - they are *all up for debate*. Right now I see that the code is rather mixed, because out of the box prettier is not really enforcing this.

Especially rules like `singleQuote: true` and `trailingComma: 'all'` are super useful.

brandones:
> Right now I see that the code is rather mixed, because out of the box prettier is not really enforcing this.

Could you provide some examples? I've never noticed anything like that. Though that's probably because I don't really worry about formatting. I don't have to think about it, which is the point of prettier.

By the same token, I think it's fine if different repos have different formatting rules — as long as they're using auto-formatting (e.g. prettier), I don't have to think about it or care.

And if someday someone decides they really don't want to use prettier in their ESM, that's probably fine. I will help encourage them to reconsider, but I'm not inclined to tell them it's against the rules.

For a similar reason, I'm not inclined to support using prettier config files. Prettier evolves our formatting according to their well-researched understanding of best practices. I'd prefer to just let them do their job.

But yes, let's definitely add `.editorconfig`s to any ESMs that are missing an `.editorconfig`.

internettrans (commented):
I disagree on this one - I think that using prettier is a great way to avoid a lot of unimportant code review back-and-forth, but don't think that every openmrs-related repo needs to arrive at a single set of code formatting rules in order to be successful.

This is a controversial topic, of course, and I expect others to have lively opinions on it :). For me, the win is using prettier at all - the specifics of which rules are enabled or not does not matter as much as whether prettier is enabled.

nkimaina:
I would vouch for this, mainly for one reason: Have you ever tried to review code from someone who is using a different style of formatting from everyone else, and they formatted a whole bunch of files? I found it difficult to tell actual changes from formatting changes, and it makes reviews very annoying.

nkimaina (approved):

drizzentic (approved):

FlorianRappl:
I think these are all great points. My motivation was / is not to force prettier on everything that goes into the final application (in any distribution), but rather to have a consistent style in all repositories maintained by the microfrontend squad.

We can, of course, reduce it to `.editorconfig` if we don't find a majority for this one. In this case I would recommend the described prettier config file (or any other - open for wishes), but it would not be enforced by this RFC.

Thoughts?

PR #25, opened 2020-04-27
First draft of 0025-webpack.md

FlorianRappl:
Thoughts and ideas for simplifying and improving our build setup per repository used for the microfrontends.

Up for discussion :rocket:!

### DIFF
--- a/text/0025-webpack.md
+++ b/text/0025-webpack.md
@@ -1,0 +1,159 @@
+# Project title
+- Start Date: 2020/04/25
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/25
+
+## Decision, including impact on distributions
+All microfrontends in the OpenMRS space should use the same base webpack
+configuration and tooling. To simplify sharing this information a common
+utility package (e.g., `@openmrs/esm-devtools`, reusing the existing package
+for everything related to development would work, too) should be created and
+used.
+
+Ideally, the new package exports commands for debugging and building. As a
+result, the *package.json* of the existing microfrontend repositories changes
+from
+
+```json
+{
+  "scripts": {
+    "start": "webpack-dev-server",
+    "build": "webpack --mode production"
+  },
+  // ...
+}
+```
+
+to
+
+```json
+{
+  "scripts": {
+    "start": "oesm debug",
+    "build": "oesm build"
+  },
+  // ...
+}
+```
+
+The rest remains the same.
+
+The commands would just use a predefined version of `webpack` and associated
+packages (e.g., `webpack-dev-server`) to trigger a build using a preconfigured
+version of a webpack file that reads the remaining information from the local
+*package.json*.
+
+The basic *webpack.config.js* would is similar to the following:
+
+```js
+const path = require("path");
+const CleanWebpackPlugin = require("clean-webpack-plugin").CleanWebpackPlugin;
+const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
+
+const root = process.cwd();
+const name = require(path.resolve(root, 'package.json')).name;
+const fileName = name.replace(/@/g, '').replace(/\//g, '-');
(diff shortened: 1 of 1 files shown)


brandones:
Yeah, this is a great idea. Also for babel config, jest config, and tsconfig. See https://github.com/openmrs/openmrs-frontend-generators/pull/1#issuecomment-604066025 . I also think it would be great to factor out the three common library mocks (api, devtools, and module-config) into their own library.

I'd factor out into something that can be installed as devDependencies. `devtools` is a `dependency` because it is webpack `external`.

I'm not sure we need a RFC for this, I think we can just do it :)  And if someday someone makes a well-reasoned decision not to use any of these things, that's probably fine.

brandones:
Right, I vote we close this and just do it? I don't think it's something that we need everyone's buy-in for.

FlorianRappl:
> Right, I vote we close this and just do it? I don't think it's something that we need everyone's buy-in for.

That means we just merge it?

I am confused here about the process, because I thought RFCs are here to bring everyone on the same page.

The same reason applies to the README PR (#23), which I think would have been valid to be merged in, too.

brandones:
> I thought RFCs are here to bring everyone on the same page.

Hmm, yeah, maybe this is a reasonable way to use them. I've seen them as being more for 1. big archetectrual decisions and 2. making rules that we intend to enforce. I don't think this is a big architectural decision, nor is it something we should enforce, per se. But on looking again I see that that's totally consistent with the wording of this RFC, assuming we read "should" as [SHOULD](https://www.ietf.org/rfc/rfc2119.txt). So yeah, this works for me.

internettrans (commented):
I agree that there is a lot of benefit to this. A couple quick notes:

1. Not every microfrontend must use webpack on a technical level - rollup, snowpack, jspm, parceljs, etc are all alternatives that are viable. This could be called out in the proposal.
2. [webpack-config-single-spa-ts](https://single-spa.js.org/docs/create-single-spa#webpack-config-single-spa-ts) is a fairly well maintained shareable webpack config that could be used to help create a shareable openmrs webpack config.
3. The `oesm` CLI referred to in this proposal actually could be rather hard to build and maintain well, and is not the approach I'd recommend. create-react-app is the example here - having its own CLI and making the webpack config buried in the node_modules makes it really hard for organizations to make necessary changes to their webpack config without "ejecting," which is notoriously pretty awful. An alternative approach is to extend the create-single-spa CLI's approach, which allows each project to have a default config that they can modify via [webpack-merge](https://github.com/survivejs/webpack-merge) or [webpack-chain](https://github.com/neutrinojs/webpack-chain). The create-single-spa CLI is designed to be extended, as [mentioned here](https://single-spa.js.org/docs/create-single-spa#generator-single-spa).
4. How does `oesm` relate to https://github.com/openmrs/openmrs-frontend-generators?

nkimaina:
I agree with @joeldenning on point 3). I am no expert in web-pack, but I have found it inevitable to once in a while modify a few things in webpack config. Microfrontends is a new concept, and I think it will be inevitable to re-configure webpack from time to time. I am in favour of any approach that does not limit my ability to customise the config.  
Given that at this point we are invested in Single SPA, then I would try to use the best practices recommended in Single SPA.

FlorianRappl:
I think these are all great points.

1. Sure, however, thinking about the microfrontend squad we should try to settle on one option.
2. I think this is a great start!
3. The CLI is nothing more than a package that comes with a good webpack config (either using  webpack-config-single-spa-ts or using the `create-single-spa` approach). Actually, the `create-react-app` is a rather bad example as they did not even provide the ability to override parts of the webpack configuration - which is really easy to do; you already mentioned some options here.
4. Actually, they will be more or less one and the same thing, i.e., using the same repository. I was not aware of this repository or that this idea was out there. Obviously, it is not yet used - but we will pick it up in this context.

> Given that at this point we are invested in Single SPA, then I would try to use the best practices recommended in Single SPA.

Fully agreed!

PR #26, opened 2020-04-28
First draft of 0026-activation-distribution.md

FlorianRappl:


brandones:
Cool, this is a great start. Thanks Florian. I suggest the following changes to the architecture you propose:

### Base

1. Create a new module that coordinates everything called `@openmrs/esm-root` (instead of using `@openmrs/esm-root-config` for it). But this is just about naming. The idea is that it no longer holds any configuration.
2. Instead of having MFEs bootstrap themselves, just have them export their own activator function (in addition to exporting its Root component). `@openmrs/esm-root` then looks makes the `registerApplication` call using that activator function.
3. This allows us to define a mechanism for overriding activator functions. Here are a few options.

### Activator function override mechanism ideas

1. Prior to registering anything, `esm-root` optionally imports something called `root-config` via the import map. The root config may export an object called something like `activatorOverrides` which looks something like
    ```
    {
      "@openmrs/esm-login": (url) => return url.includes("loggy-inny")
    }
    ```
    `esm-root` would then register esm-login using that custom activator rather than the one esm-login provides for itself.

2. Prior to registering anything, `esm-root` optionally imports `config-file` from the import map, the same one used by `module-config`. This is generally a JSON file. `esm-root` looks in the object it receives for a structure like
    ```
    {
      "root": {
        "@openmrs/esm-login": "url.includes('loggy-inny')"
      },
      ...
    ```
    `esm-root` uses `eval` to create the activator function override.

The advantages of the first option are 1. provides a way for implementations to execute code prior to starting the application (like a custom root config does normally), 2. no JS in a JSON file, and 3. no eval.

The advantage of the second option is that it keeps things simple. There's only one file needed to configure OpenMRS SPA, and it's more or less in JSON. The importance of this shouldn't be understated—most people find the "root config vs module config" distinction very confusing.

Note that these approaches are totally compatible -- the latter could be "the easy/common way" to provide activator overrides, and the former could be used by those who need fancier overrides or want to e.g. provide a hierarchy of module config files.

How does all that sound?

FlorianRappl:
Hi @brandones - great proposals.

The only small change that I see is the following:
I would not create a new module, but rather include it in the initial script ("app shell"). Right now the initial script only comes with SystemJS and Single-SPA - it should have this functionality, too.

Regarding the export I would propose to use

```js
export function setup() {
  return {
    activate: /*activator function */,
    lifecycle: /*resolver for lifecycle */,
  };
}
```

so we'd only need to check if the module (i.e., microfrontend root module) exports a `setup` function.

For the activator override I think both (1) and (2) make sense. If I would need to choose (e.g., for an initial implementation) I would choose (1). Reason is that overriding the activator should not be required at all, so if it is required being more powerful (i.e., choosing (1)) could be demanded right away.

What are your thoughts?

brandones:
Sorry, having a bit of trouble wrapping my brain around this...

> I would not create a new module, but rather include it in the initial script ("app shell"). Right now the initial script only comes with SystemJS and Single-SPA - it should have this functionality, too.

What's the initial script, and how does it get called?

> Regarding the export I would propose to use
> 
> ```js
> export function setup() {
>   return {
>     activate: /*activator function */,
>     lifecycle: /*resolver for lifecycle */,
>   };
> }
> ```
> 
> so we'd only need to check if the module (i.e., microfrontend root module) exports a `setup` function.

This is the root-config, right? Not the normal MFEs? What do `activate` and `lifecycle` do?
 
> For the activator override I think both (1) and (2) make sense. If I would need to choose (e.g., for an initial implementation) I would choose (1). Reason is that overriding the activator should not be required at all, so if it is required being more powerful (i.e., choosing (1)) could be demanded right away.

Yes, that makes perfect sense to me.

FlorianRappl:
The initial script is:

```js
window.openmrsBase = "/openmrs";
window.spaBase = "/spa";
window.getOpenmrsSpaBase = function() { return window.openmrsBase + window.spaBase + '/';};
System.import('@openmrs/esm-root-config');
System.import('@openmrs/esm-styleguide');
```

It sets up important globals and imports the important global modules.

(actually to be fair this script should also be minified etc. and combined with a couple of other resources, see below,

```html
<script src="/openmrs/moduleResources/spa/node_modules/import-map-overrides/dist/import-map-overrides.js"></script>
<script src="/openmrs/moduleResources/spa/node_modules/systemjs/dist/system.min.js"></script>
<script src="/openmrs/moduleResources/spa/node_modules/systemjs/dist/extras/amd.min.js"></script>
<script src="/openmrs/moduleResources/spa/node_modules/systemjs/dist/extras/named-exports.min.js"></script>
<script src="/openmrs/moduleResources/spa/node_modules/systemjs/dist/extras/named-register.min.js"></script>
<script src="/openmrs/moduleResources/spa/node_modules/systemjs/dist/extras/use-default.min.js"></script>
```

but that is not so important for the discussion.)

This script gets called implicitly because it is part of the app shell (hence initial script). The app shell is the central app (right now https://openmrs-spa.org/openmrs/spa/).

> What do activate and lifecycle do?

Activate is the activation function and lifecycle is essentially the `LifeCycles<T> | ((config: T & AppProps) => Promise<LifeCycles<T> | Splat<LifeCycles<T>>>)` as you know it (expectation of `registerApplication`).

brandones:
Regarding the initial script -- oh yeah, that makes perfect sense to me!

Regarding the exports:

It makes sense to me that each MFE might export something like
```js
export function setup() {
  return {
    activate: /*activator function */,
    lifecycle: /*resolver for lifecycle */,
  };
}
```

which specifies the activator and lifecycle for that MFE. But it seems like you're proposing that structure for the root config? If that's the case, it doesn't seem adequate. The root config's job is to optionally override the activator function for *any* module. It doesn't have an activator function or a lifecycle -- the app shell just knows what to do with it. So maybe something vaguely like
```js
// app shell
const activators = getActivatorsFromTheirModulesSetupFunctions();
try {
  const { activatorOverrides } = await System.import("root-config");
  activators.assign(activatorOverrides);
} catch (Exception e) {
  // no root config, no problem
}
for (let m of Object.keys(activators)) {
  registerApplication(m, applications[m], activators[m]);
}
```

```js
// root-config
export const activatorOverrides = {
  "@openmrs/esm-login": // custom esm-login activator function
}
```

Does that make sense?

FlorianRappl:
> But it seems like you're proposing that structure for the root config?

No the structure is only there for the microfrontends, not library modules (e.g., React does not come with such a structure; nor does any other external library => we should also not expect internal libraries to have this structure).

The root config also does not register itself as an application, right?

I think we are at the same page - the two parts are conceptionally as you outlined.

Just to reflect on what I wrote initially:

> [...] if the module (i.e., microfrontend root module) exports a setup function.

A microfrontend root module is, e.g., the `index.ts` that right now has the call to the `registerApplication` of Single SPA. It is the entry point for Webpack (thus "root"). Sorry if there was confusion with "root config".

brandones:
Yes! Okay, I think we are indeed on the same page.

Right, the root config does not register itself as an application.

brandones:
So I was drafting a recommendation for how to rework this, but I realized there's an unsolved problem we need to address:

How does this work for cloud-based distributions, such as openmrs-spa.org ? What's cool about the current architecture is that they don't need to change anything on the server-side to add or remove modules.

Two ideas:

1. We add a Global Property that tells the server whether to use the server-generated app array (what you call `window.spaApps`) or whether to get the app array from a URL. If using the server-generated app array, the server will embed the array in the `index.html` file (as in your example), based on the MFEs that the server knows about. If using a URL, it will just use whatever list of apps is provided at that URL, regardless of what's installed server-side.

2. We assume that everything in the import map should be loaded. The app shell loads everything in the import map. For each bundle loaded, if that bundle satisfies the MFE API (i.e. it exports `activator` and `lifecycle`), the app shell registers it with Single-SPA. Otherwise the app shell ignores it.

I lean toward option (2).

FlorianRappl:
> there's an unsolved problem we need to address

Actually no - that problem is solved already, but we can still use 1) or 2).

My reason for avoiding 2) was that I had the impression we did not want to load all libraries. They should actually be lazy loaded. If that is not the case, then sure going for 2) - it would simplify things.

Now regarding 1) I think the property of the package.json was not really considered. This is how the server recognizes a MFE. I don't know how the upload is currently done, but you can then either change the POST endpoint to not only take a URL, but also an optional flag if its a MFE, or - for offline purposes - take an NPM package / tarball that contains all necessary files (incl. the package.json). The info from the package.json now determines if its a MFE.

> What's cool about the current architecture is that they don't need to change anything on the server-side to add or remove modules

Well, actually it needs to know, right? The importmap is served by the server and the root config has the other info. So changing anything requires multiple changes according to that logic (that's one of the reasons for this PR in the first place). With the new change the server still computes the importmap, but also just places the MFEs. So these are not fixed and they still work with the cloud approach.

As written performing updates (this not only includes updating existing modules, but also adding new ones) should be possible. The approach took both ways (online and offline) into account.

Bottom line: If we agree that lazy loading of deps in the importmap is not wanted then let's just go for 2).

brandones:
Yeah, I don't think lazy loading of deps in the import map will be needed. Non-MFEs that are in the import map are there because 1. they are used by almost every module or 2. they need to share one instance/state across the application. It seems unlikely that someone will want to lazy-load a module that meets either of those two criteria.

If someone disagrees with me we can get into the weeds of option 1 :)

brandones:
Okay, I think we're conceptually in good shape. Here are some changes I'd like to see to the document.

I think "definitions" should include the following (these are just some proposed names and sketched definitions):

- **root config** *Replaces the old concept of a [root config](https://wiki.openmrs.org/display/projects/openmrs-esm-root-config)*. An optional element of the import map, named `root-config`, which can be used to 1. provide module configs to esm-module-config, 2. override module activator functions, and 3. do any other pre-app-launch setup the implementer wants. It *does not* register microfrontends with Single-SPA.
- **microfrontend interface** conditions on the the exports of an ESM which, if met, indicate that the module is an OpenMRS Microfrontend.

Maybe with bits of example code or pseudocode. The "decision" then becomes something like

*OpenMRS Microfrontends will have their own default activator functions. Microfrontends will be identified as such by their exports satisfying a specific microfrontend interface, defined in this RFC. The app shell will load the app shell config if available. The app shell will register all microfrontends in the import map with Single-SPA, and start the application.

The existing concept of a "root config" will be superseded by the app shell and app shell config.

In order to implement this change, we will need to
1. Add `index.ts` entrypoint files to each microfrontend that satisfy the microfrontend interface (exporting `lifecycle` and `activate`).
2. Change the initial script to behave as described.
3. Migrate all existing root configs to the new structure. This will generally just mean removing the calls to `registerApplication` and exporting those activators that are different from the defaults.*

And then everything else goes in "Reasons" -- "properly distributing" microfrontends, making autonomous and easier to develop, easier to create a distro, etc.

Does this all sound right and reasonable?

brandones:
Also, I just want to make sure that we're on the same page that

>The `index.ts` may be as simple as follows:
>
> ```js
> // set the public path to be able to retrieve side bundles
> import "./set-public-path";
> // register the application in the shared single-spa package
> import { registerApplication } from "single-spa";
> // get useful helpers for the activity functions from the loaded root-config
> import { routePrefix } from '@openmrs/esm-root-config';
> // perform the registration with the lazy loading of the lifecycle and its activity function
> registerApplication(
>   "@openmrs/esm-home",
>   () => import('./openmrs-esm-home'),
>   location => routePrefix("home", location),
> );
> // export other dependencies to be used, e.g., by the dev tooling
> export { backendDependencies } from "./openmrs-backend-dependencies";
> ```

No longer applies; that `index.ts` should satisfy the interface that you call `setupOpenMRS` and I call the "microfrontend interface," and it should not call `registerApplication`.

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,102 @@
...
+// perform the registration with the lazy loading of the lifecycle and its activity function
+registerApplication(
+  "@openmrs/esm-home",
+  () => import('./openmrs-esm-home'),
+  location => routePrefix("home", location),
+);
+
+// export other dependencies to be used, e.g., by the dev tooling
+export { backendDependencies } from "./openmrs-backend-dependencies";
+```
```
brandones: Also, I just want to make sure that we're on the same page that this no longer applies; that index.ts should satisfy the interface that you call setupOpenMRS and I call the "microfrontend interface," and it should not call registerApplication. Yeah?

FlorianRappl: Good catch! I'll correct it!

brandones:
Also, at this point I'd like to ask @joeldenning , if you have the time, I'd really value your input on this. We're talking about making some changes to how the application and microfrontends get loaded.

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+- Improve developer experience
+
+## Alternatives
+There are three main alternatives that can be seen:
+
+1. Switching away from Single SPA to microfrontend frameworks that support the
+notion of distributed microfrontends.
+2. Keeping the original notion of a single file that has the ultimate knowledge
+(with all the drawbacks and complexity for distribution).
+3. Using a service to get the information via an API call. Besides the original
```
internettrans, line 107: As of single-spa@5.3.0, single-spa accepts strings as activity functions. For example:

```js
singleSpa.registerApplication({
  name: 'nav',
  app: () => System.import('nav'),
  activeWhen: ['/']
})

singleSpa.registerApplication({
  name: 'patient-chart',
  app: () => System.import('patient-chart'),
  activeWhen: ['/patient/:uuid/chart']
})
```

- [Release notes](https://github.com/single-spa/single-spa/releases/tag/v5.3.0)
- [Documentation](https://single-spa.js.org/docs/configuration#configactivewhen)

^ I think that storing a root config in a service/database is a very viable approach, and is one that I know several organizations are already employing. The database could store the name of the service and a list of activeWhen strings - that is all that is needed to register the application.

internettrans (commented):
Hey there 👋 I'm glad to see that this RFC is still useful and that there are new contributors to it! I think the proposal has some good aims, but have some concerns about the approach. I think I lean toward Alternative 3, where root config information is retrieved from a service that uses a database to manage the applications.

### Main Concern

If I understand this proposal correctly, I think that it would require downloading the code for all microfrontends in order to register the single-spa applications? If so, that comes with a pretty severe performance penalty that I think is undesireable and avoidable. Lazy loading the microfrontends is crucial to a fast time-to-first-render.

When I originally authored and designed single-spa, I intentionally moved the activity and loading functions into the root config so as to allow for lazy loading of the individual microfrontends.

Please let me know if I've misunderstood something about the proposal. I didn't read through the entire discussion above, but did read through the whole proposal text. I might have missed something 😄

### Alternative solution

I think your proposed Alternative 3 would be better for this. Other organizations that use single-spa ([1](https://github.com/single-spa/single-spa/issues/525#issuecomment-619374560), [2](https://github.com/single-spa/single-spa/issues/525#issuecomment-619265340)) store their root config in a database as json, embed the json into the HTML file, and dynamically call registerApplication with that json.

You brought up a drawback of Alternative 3 that is not necessarily true and can be worked around pretty easily. I commented below on that with example code that shows how. The short explanation is that as of single-spa@5.3.0, you may now pass in a simple string as an activity function, which single-spa will convert into a full activity function.

### Something else to consider

Due to some feedback from several members of the single-spa community, I am actively working on a new npm package called `single-spa-layout` which will allow for dynamic creation of single-spa applications based on either JSON or HTMLElements. We have an RFC for single-spa going on at https://github.com/single-spa/single-spa/issues/525, and I'd love for the OpenMRS community to chime and provide feedback. If OpenMRS does end up with storing root configs inside of a database, I think it might make sense to consider storing the JSON used by single-spa-layout instead of the raw application objects.

text/0026-activation-distribution.md
FlorianRappl: > If I understand this proposal correctly, I think that it would require downloading the code for all microfrontends in order to register the single-spa applications?

No - the microfrontends are bundle splitted as stated; so you only download essentially 0.5 KB - not the whole MF. So I assume this was misunderstood.

> As of single-spa@5.3.0, single-spa accepts strings as activity functions.

That's great and certainly useful also for the JSON configuration. However, it does not help with the distribution.

> I think I lean toward Alternative 3

Wonderful - thanks for the suggestion! :)

internettrans: > No - the microfrontends are bundle splitted as stated; so you only download essentially 0.5 KB - not the whole MF.

I've read through the proposal several times and do not see where it discusses code splitting. Could you help me understand this? Even with code splitting, this still comes with a performance cost - if there are 25 microfrontends, this proposal requires downloading the entry bundle for all 25 of them in order to register the applications?

internettrans: > That's great and certainly useful also for the JSON configuration. However, it does not help with the distribution.

Could you help me understand this? Distributions already have a variety of configuration stored in the database that is accessible via the Admin tool. I believe that this includes which Java modules are activated or not, which is fairly similar to which javascript modules are activated. I agree that storing code in a database has downsides, but don't see how it prevents distribution, since my understanding of creating OpenMRS distributions is that this often involves setting up a database with correct config.

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+```js
+// set the public path to be able to retrieve side bundles
+import "./set-public-path";
+// get useful helpers for the activity functions from the loaded root-config
+import { routePrefix } from '@openmrs/esm-root-config';
+
+export function setupOpenMRS() {
+  return {
+    lifecycle: () => import('./openmrs-esm-home'),
```
FlorianRappl, line 83: @joeldenning -> side-bundle.

text/0026-activation-distribution.md
FlorianRappl: > if there are 25 microfrontends, this proposal requires downloading the entry bundle for all 25 of them in order to register the applications?

Right now we have multiple loose scripts via importmaps, which could also be served as a single UMD. You cannot argue in favor of fixed importmaps (+ include 6 loose SystemJS scripts, all which could be served as a single UMD, too), but not be in favor of microfrontends being loosely coupled. Microfrontends are all about loose coupling.

Furthermore, this only needs to happen once. The script can then be cached.

I am not sure if I read all that correctly - to me it reads with a certain attitude (maybe I am mistaking here). I propose to discuss this in a call. I guess it would make more sense to you then.

internettrans: My intent here is engaging in a respectful, productive dialog. I am not arguing with you, but seeking to understand. Thanks for the [clarification comment](https://github.com/openmrs/openmrs-rfc-frontend/pull/26#discussion_r420865829) below calling out the code split via `import()`, that helps me understand how this proposal works 👍.

> You cannot argue in favor of fixed importmaps (+ include 6 loose SystemJS scripts, all which could be served as a single UMD, too), but not be in favor of microfrontends being loosely coupled. Microfrontends are all about loose coupling.

I don't view putting top level routes into a root config as tight coupling. And do not see the connection between loading SystemJS scripts and the current discussion here.

Moving the activity function logic into the microfrontend itself does make it more self encapsulated. But I think that we shouldn't sacrifice performance for that encapsulation.

#### Summary of my concern

The current proposal requires that every microfrontend's entry bundle be downloaded before the first meaningful render. The entry bundles can be quite small due to code splits, but doing this is still is a fairly large and dynamic number of network requests before the first meaningful render. I think that a fairly large, dynamic number of network requests has a performance penalty that should be avoided if possible.

#### Maybe another alternative?

Perhaps instead of downloading all modules in the browser in order to retrieve their activity functions, each package could publish to npm a separate file that could be included into the root-config as a build-time dependency?

Example:

```js
// root-config.js

// these are "build-time" modules that come from node_modules
import registerPatientChart from '@openmrs/esm-patient-chart/registration'
import registerLogin from '@openmrs/esm-login/registration'

registerPatientChart()
registerLogin()
```

```js
// patient-chart/dist/registration.js
import { registerApplication } from 'single-spa';

export default function () {
  registerApplication({
    name: '@openmrs/esm-patient-chart',
    app: () => System.import('@openmrs/esm-patient-chart'),
    activeWhen: ['/patient/:uuid/chart']
  })
}
```

FlorianRappl: Let me try to show you the connection to SystemJS (which is just one example). You refer about saving requests, which obviously contradict the current design (e.g., instead of a single bundle we serve 7 scripts, 6 references, 1 inline). I would love to set the same metrics everywhere.

> each package could publish to npm a separate file that could be included into the root-config as a build-time dependency?

Here I think there is a misunderstanding. You have several independently developed and deployed modules. How can they now come together in a build (coming from a system that is shipped pre-built and does not have any knowledge of which and how many modules are actually used later on)?

The idea of this proposal is to allow a free composition.

internettrans: > You refer about saving requests, which obviously contradict the current design (e.g., instead of a single bundle we serve 7 scripts, 6 references, 1 inline). I would love to set the same metrics everywhere.

The difference between the systemjs scripts and this proposal is that the systemjs scripts are a fixed 5-6 requests, whereas the number of applications is dynamic, and could be as high as 25 or even 50 (for a large organization). The systemjs library itself is quite small (~5kb gzipped for everything).

A dynamic number of requests has a lot more potential to impact performance than a fixed number of requests.

> Here I think there is a misunderstanding. You have several independently developed and deployed modules. How can they now come together in a build (coming from a system that is shipped pre-built and does not have any knowledge of which and how many modules are actually used later on)?

Most OpenMRS distributions do not independently deploy modules, but rather package them all up and ship them out as `.war` files. In order for deployments to be possible, there must be centralization of servers, which is not true of the majority of the OpenMRS community. As far as I know, there are only very few organizations (Ampath being one of them) that have centralized their servers.

I agree that my proposal disallows for independent deployment, which is not ideal. That's a valid criticism and perhaps the idea should be scrapped 😄.

My intent here is to try to find a way to avoid an unknown and dynamic number of network requests before the first meaningful render.

FlorianRappl: > My intent here is to try to find a way to avoid an unknown and dynamic number of network requests before the first meaningful render.

I think this is good and obviously the argument with static vs dynamic number is also true. Nevertheless, the argument that these initial requests should be already bundled is in the same sense also right - and justifies that we are not playing with the same rules here. Unfortunately.

Nevertheless, coming back to the topic. The thinking with the unknown number of requests does not take into consideration HTTP/2 batching and multiplexing. The cost of multiple requests for scripts that are ~less than 1 KB in size is in reality quite negligible. And again - that these requests exist is only true for FTU. Afterwards, they are all there - which by the way also improves update scenarios. Instead of two things you'll only need to update one module. In this case only the updated single script is requested from the browser.

internettrans: Fair enough - if it is ensured that the code splitting occurs, http2 enabled, gzip, etc, perhaps the performance impact is worth the benefits of moving the top level routes to the microfrontends themselves.

I'll leave this conversation here for now and let the MF core team evaluate this PR further. I am not as involved in openmrs as I was last year and view my role here as just providing discussion points / insight, but not as the decision maker.

I think you're a smart person @FlorianRappl who clearly has some experience and knows your stuff. I'm glad OpenMRS has you to help push things forward.

brandones:
Lots of interesting ideas above. Thanks so much for having taken the time to help us think this through, @joeldenning . I'll try and synthesize and move things forward as best I can.

@FlorianRappl , I think we need to take the concern about performance seriously, because OpenMRS is very often run on old hardware and unreliable networks. At the same time, what I'm reading suggests that you should be right about HTTP/2 making the performance hit trivial. I'm happy to be optimistic about HTTP/2 batching and multiplexing being enough, but I'd still rather not foreclose on the possibility of needing the server know the activators ahead of time.

I think the proposal as it stands supports that. I would, however, like to suggest another alternative, which might make that easier (and has some other advantages too, I think).

Instead of using dynamic code splitting, we could use entry point splitting. Each MFE then has two different imports available, e.g. `import "@openmrs/esm-login"`, which works exactly the same as it does now, and `import "@openmrs/esm-login/registration"`, which is an object defining an activator function.

In order to support package imports (like `@openmrs/esm-login/registration`), we would normally have to [add some lines](https://github.com/WICG/import-maps#packages-via-trailing-slashes) to the import map. However, because import maps support [dynamic generation](https://github.com/WICG/import-maps#dynamic-import-map-example), we could have the package import entries inferred from the normal entries.

So the initial script does the following:
1. Generates the package import line for everything in the import map
2. Attempts to get the "registration" script for every package
3. If a registration file doesn't exist, ignore it
4. If a registration file does exist, register the ESM using the activator function it provides

Advantages:
1. Doesn't require downloading everything in the import map. There is still one call for each item, but all the non-MFEs will 404. I assume this would be faster than downloading everything in the import map (granting that all the MFEs are code-split).
2. Compatibility with the normal Single-SPA ecosystem. Backward-compatibility.
3. It's more obvious to me how one would implement server-side caching of the registration script. I don't know enough about how dynamic code splits work with import maps to know if caching the entry bundle would be easy, hard, or impossible. But it would be very straightforward with an entrypoint-split bundle.

Of course, if we end up having to implement server-side caching, we then have a cache invalidation problem. I'd note, however, that it's equivalent to the database updating problem presented by Alternative 3. However we imagine we'd keep the database's cache of activator functions updated, we can handle this cache in the same way.

Interested in hearing what you think!

FlorianRappl:
I am not sure this addresses the performance concern. You'd still have a lot of requests. If the concern is about the import map - well, according to the original proposal it can still be lazy loaded (it was your suggestion to remove lazy loading here).

Advantage (2) I do not really understand - this is all still standard Single SPA. Single SPA is just a small library, there is nothing to tell you how to split up your application or how to get these activator functions.

In the past days I've been actively working with new collaborators on new microfrontends for OpenMRS. As it is the workflow is way too complicated and blocking. They need to run 3 (!) dev servers, make multiple modifications, and they need to know about changing / debugging the reference site. Microfrontends should be the opposite. This proposal is just one step towards that goal.

brandones:
> I am not sure this addresses the performance concern. You'd still have a lot of requests.

Yes, the number of requests would be the same. However we would no longer need to actually download everything in the import map. We would succeed in lazy loading everything but the registration bundles.

> If the concern is about the import map - well, according to the original proposal it can still be lazy loaded (it was your suggestion to remove lazy loading here).

Sure, I mean, the same technique could be applied to this proposal -- if the server has an explicit list of MFEs it can download registration bundles only for those and avoid the 404s. But I think it's better if we can avoid needing to manage such a list.

> Advantage (2) I do not really understand - this is all still standard Single SPA. Single SPA is just a small library, there is nothing to tell you how to split up your application or how to get these activator functions.

Yes, very good point.

> In the past days I've been actively working with new collaborators on new microfrontends for OpenMRS. As it is the workflow is way too complicated and blocking. They need to run 3 (!) dev servers, make multiple modifications, and they need to know about changing / debugging the reference site. Microfrontends should be the opposite. This proposal is just one step towards that goal.

Yeah! I think we're in strong agreement that this should happen. We're just hashing out a technical detail about your proposal -- whether we should use dynamic splitting or entrypoint splitting to achieve what we all want to achieve. I'm not married to either, but I think we should give each their due consideration.

FlorianRappl:
Fully agreed.

I would then suggest we take it by convention. If a module ends with `-app` (other suggestions appreciated) we'll load it right away. This implies, e.g., renaming `@openmrs/esm-patient-chart` to `@openmrs/esm-patient-chart-app`.

brandones:
Wow, that's so simple and extremely sensible I almost can't believe we didn't think of it before. Yes. Good idea.

The argument that remains in favor of entrypoint splitting over dynamic splitting is (3), that it would be easy to cache entrypoint-split registration bundles server-side, whereas I don't know if the same is true of dynamically-split bundles.

FlorianRappl:
Not sure where the problem is with that server-side caching. Usually, you would send back the script and cache it indefinitely on the client. Imho we should just put it in a directory with the current version (or hash) and then there is no problem in either case.

As an example today we get

```
https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-primary-navigation/e0059ac1749253cf776933fd3c661c92f58ecfed/openmrs-esm-primary-navigation.js
```

for one of the modules. Therefore, the dynamically split bundle is potentially at

```
https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-primary-navigation/e0059ac1749253cf776933fd3c661c92f58ecfed/0.index.js
```

(or something similar; naming up to webpack config, but you get the idea)

Where is this not cachable or less cachable? It can just be cached indefinitely. The `e0059ac1749253cf776933fd3c661c92f58ecfed` is (I guess, at least it should be) the hash of the respective release. Having a new release will result in a new entry in the import map and thus the dynamically split bundle just sits in the new same directory (i.e., is not yet cached, but will be after downloading it). Maybe I am missing something here.

brandones:
Right, so the difference between client-side and server-side caching here is important. Yeah, with client side caching, everything works as expected.

Server-side caching is different, because normally the server doesn't ever fetch the MFEs at all. If we want to implement server-side caching of registration bundles, the server will have to know how to access the registration bundle. How does the server know where to find the registration bundle? Also, that registration bundle dynamically imports the application bundle. If the server passes the cached registration bundle embedded in the init script, will the client know where to find the application bundle?

But maybe all this is moot, maybe we should just bank on not having to do server-side caching at all. It probably wouldn't be worth the additional complexity.

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
+# Project title
+- Start Date: 2020/04/27
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/26
+
+## Decision, including impact on distributions
+OpenMRS Microfrontends will have their own default activator functions.
+Microfrontends will be identified as such by their exports satisfying a
+specific microfrontend interface, defined in this RFC. The app shell will load
+the app shell config if available. The app shell will register all
+microfrontends in the import map with Single SPA, and start the application.
```
brandones:
```suggestion
Microfrontends will be identified as such by their names being suffixed with `-app`.
Microfrontends' will provide activator and lifecycle functions in their exports according
to a specific microfrontend interface, defined in  this RFC. The interface bundle will be
dynamically split from the main application bundle via a dynamic import in the lifecycle function.
The app shell will load the app shell config if available. The app shell will register all
microfrontends in the import map with Single SPA, and start the application.
```

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+1. Add index.ts entrypoint files to each microfrontend that satisfy the
+microfrontend interface (exporting lifecycle and activate).
+2. Change the initial script to behave as described.
+3. Migrate all existing root configs to the new structure. This will generally
+just mean removing the calls to registerApplication and exporting those
+activators that are different from the defaults.
+
+Instead of lazy loading all modules from the importmaps the modules are
+directly evaluated. 
+
```
brandones:
```suggestion
```

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+An **app shell** refers to the index.html page that includes the required base
+scripts. Thus triggering the whole loading process necessary for all the
+microfrontends.
+
+The **initial script** is the script performing the loading of the Single SPA
+library. It sets up important globals and imports the important global modules.
+
+The new **root config** replaces the old concept of a [root config][1]. An
+optional element of the import map, named `@openmrs/esm-root-config`, which can
```
brandones:
```suggestion
optional element of the import map, named `root-config`, which can
```

My understanding is that `@openmrs/esm-root-config` will either cease to exist or be converted into the *initial script*, right? This new "root config" (maybe we should call it something else?) will be totally different. There will be no default version of it, it's totally optional and implementation-specific. So it wouldn't make sense to namespace it to `@openmrs`. Does that make sense?

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+The **lifecycle** of an app determines what to do technically for booting,
+mounting, updating, and unmounting an application in the DOM of the currently
+presented website.
+
+The **DOM** refers to the API responsible for the object representation of a
+website in the browser.
+
+An **app shell** refers to the index.html page that includes the required base
+scripts. Thus triggering the whole loading process necessary for all the
+microfrontends.
```
brandones:
```suggestion
An **app shell** refers to the index.html page that includes required libraries
and the *initial script*.
```

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+The **initial script** is the script performing the loading of the Single SPA
+library. It sets up important globals and imports the important global modules.
```
brandones:
```suggestion
The **initial script** is the script which identifies microfrontends in the import map,
loads them, loads the *root config*, and registers microfrontends with the Single SPA
application. It sets up important globals and imports the important global modules.
```

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+The new **root config** replaces the old concept of a [root config][1]. An
+optional element of the import map, named `@openmrs/esm-root-config`, which can
+be used to (1) provide module configs to `@openmrs/esm-module-config`, (2)
+override module activator functions, and (3) do any other pre-app-launch setup
+wanted by the actual implemention. Removed responsibility: Registration of
+microfrontends with the Single SPA framework.
```
brandones:
```suggestion
wanted by the actual implemention. It does not register
microfrontends with the Single SPA framework.
```

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+A **microfrontend interface** describes the conditions on the exports of an ESM
+which, if met, indicate that the module is an OpenMRS microfrontend module. In
+particular, the modules exporting a `setupOpenMRS` are considered a
+microfrontend module.
+
+The interface for the `setupOpenMRS` function can be described in TypeScript as
```
brandones:
```suggestion
The **microfrontend interface** is the interface that the *initial script* expects microfrontends
to satisfy. Microfrontends are ESMs whose names are suffixed with `-app`. The interface
consists of a single function, `setupOpenMRS`:
```

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,120 @@
...
+
+```ts
+interface SetupOpenMRS {
+  (): {
+	  lifecycle: LifeCycles<T> | (() => Promise<LifeCycles<T> | Splat<LifeCycles<T>>>);
+    activate(location: Location): boolean;
```
brandones:
```suggestion
    lifecycle: LifeCycles<T> | (() => Promise<LifeCycles<T> | Splat<LifeCycles<T>>>);
    activate(location: Location): boolean;
```

brandones (changes requested):

brandones:
Cool, I've submitted some suggestions based on our discussion. Hopefully we're on the same page. Let's get some other reviewers to look this over. Thanks so much for the patience, energy, and thought you've put into all of this, @FlorianRappl .

jonathandick (approved):

mks-d (approved):

bmamlin (approved):

text/0026-activation-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+OpenMRS Microfrontends will have their own default activator functions.
+Microfrontends will be identified as such by their names being suffixed with `-app`.
+Microfrontends' will provide activator and lifecycle functions in their exports according
+to a specific microfrontend interface, defined in  this RFC. The interface bundle will be
+dynamically split from the main application bundle via a dynamic import in the lifecycle function.
+The app shell will load the app shell config if available. The app shell will register all
+microfrontends in the import map with Single SPA, and start the application.
+
+The existing concept of a "root config" will be superseded by the app shell
+and an app shell config.
```
brandones:
```suggestion
The existing concept of a "root config" will be superseded by the initial script,
which is optionally configurable by a new kind of "root config."
```

brandones: @FlorianRappl Shoot, just this one more, and then that's it!

FlorianRappl: Awesome thanks for the enhancement @brandones ! :rocket:

brandones (approved):

PR #27, opened 2020-08-22
Extensions implementation plan

brandones:
This is not a traditional RFC; it is more of an implementation plan and detailed technical description of what we are building. It is likely that it contains errors and technical impossibilities and will need some updating in the course of its implementation. It should be considered a living document for that duration. At the end we can commit it and have a detailed explanation of how all this works.

Many thanks to @FlorianRappl for providing both the groundwork and inspiration for this.

Please see ~[slides 17-end of this presentation](https://docs.google.com/presentation/d/1ParNFdehbBexycC_XzdvpPNXBCea-4GYwAuoPFtvYIY/edit#slide=id.g921ee92cfb_0_2)~ [these slides](https://docs.google.com/presentation/d/1GmogTn-YSc_TBAvRyakeaBT-7IYsvvUvzZum21LAIOg/edit?usp=sharing) for context.

Epic: https://issues.openmrs.org/browse/MF-294

### DIFF
--- a/text/0027-extensions.md
+++ b/text/0027-extensions.md
@@ -58,6 +58,11 @@ where many of the below concepts are introduced visually.
 
 ## API Proposal
 
+Applications are registered in the `setupOpenMRS` function of a module. The `pages`
+key should accept an array of objects with the following properties:
+- route: equivalent to current `activate` parameter
+- load: a function that, when called, returns an object with the lifecycle methods
+
 Extensions are registered in the `setupOpenMRS` function of a module. The `extensions`
 key should accept an array of objects with the following properties:
 - name: the unique name of the extension
@@ -67,8 +72,10 @@ key should accept an array of objects with the following properties:
 ```javascript
 function setupOpenMRS() {
   return {
-    lifecycle: () => import("./openmrs-esm-login"),
-    activate: "login",
+    pages: [
+        { route: "login",
+          load: () => import("./openmrs-esm-login") }
+    ],
     extensions: [
         { name: "locationPicker",
           type: "setting",
--- a/text/0027-extensions.md
+++ b/text/0027-extensions.md
@@ -13,10 +13,6 @@ We'll build an UI in a new module `@openmrs/esm-implementer-tools`
 for managing that configuration, which will assist implementers
 in changing the config to add, remove, or reorder extensions.
 
-All servers will have a config file `config.json` in the `frontends` directory
-of the application data directory. This file can be written to by the server,
-and managed through the UI.
-
 ## Definition
 
 An **Extension** is a component with a unique *name* and a *type*.
@@ -53,12 +49,12 @@ interface Lifecycle {
 ```
 
 Please see
-[slides 17-end of this presentation](https://docs.google.com/presentation/d/1ParNFdehbBexycC_XzdvpPNXBCea-4GYwAuoPFtvYIY/edit#slide=id.g921ee92cfb_0_2),
+[these slides](https://docs.google.com/presentation/d/1GmogTn-YSc_TBAvRyakeaBT-7IYsvvUvzZum21LAIOg/edit#slide=id.g1f88252dc4_0_162)
 where many of the below concepts are introduced visually.
 
 ## API Proposal
 
-Applications are registered in the `setupOpenMRS` function of a module. The `pages`
+Pages are registered in the `setupOpenMRS` function of a module. The `pages`
 key should accept an array of objects with the following properties:
 - route: equivalent to current `activate` parameter
 - load: a function that, when called, returns an object with the lifecycle methods
@@ -66,8 +62,10 @@ key should accept an array of objects with the following properties:
 Extensions are registered in the `setupOpenMRS` function of a module. The `extensions`
 key should accept an array of objects with the following properties:
 - name: the unique name of the extension
-- type: the autocomplete type of the extension
 - load: a function that, when called, returns an object with the lifecycle methods
+- type: [optional] the autocomplete type of the extension
+- slot: [optional] the name of a slot to attach to
+- slots: [optional] an array of slots to attach to
 
 ```javascript
 function setupOpenMRS() {
@@ -79,6 +77,7 @@ function setupOpenMRS() {
     extensions: [
         { name: "locationPicker",
           type: "setting",
+          slot: "
           load: () => import("./location-picker") }
     ]
   };
@@ -121,12 +120,12 @@ Extensions can be associated with extension slots programmatically by calling `a
 attach(extensionSlotName: string, extensionName: string): void
 ```
 
-This can take place either on the extension slot side, or on the exension side
-where `setupOpenMRS` is defined.
+This can take place anywhere, but will usually be either on the extension slot side,
+or on the exension side where `setupOpenMRS` is defined.
 
 If the same extension is attached to the same extension slot multiple times,
-the `extensionName` must be made unique by suffixing it with `#[extensionId]`,
-where `[extensionId]` is an identifier of the developer's choosing—preferably
+the `extensionName` must be made unique by suffixing it with `#[idSuffix]`,
+where `[idSuffix]` is an identifier of the developer's choosing—preferably
 one that is descriptive and human-friendly. So
 `notesWidget#hivNotes` would be a valid value for `extensionName`.
 
@@ -150,14 +149,14 @@ Every module that has extension slots implicitly supports configuration of the f
 
 ```js
 {
-  "my-module-name": {
-    "extensions": {
-      `extensionSlotName`: {
+  [moduleName]: {
+    "extensionSlots": {
+      [extensionSlotName]: {
         "add": // Array<string>, array of extension IDs
         "remove": // Array<string>, array of extension IDs
         "order": // Array<string>, array of extension IDs
         "configure": {
-          `extensionId`: // object
+          [extensionId]: // object
         }
       }
     }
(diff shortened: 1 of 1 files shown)


text/0027-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+- type: the autocomplete type of the extension
+- load: a function that, when called, imports the extension
+
+```javascript
+function setupOpenMRS() {
+  return {
+    lifecycle: () => import("./openmrs-esm-login"),
+    activate: "login",
+    extensions: [
+        { name: “locationPicker”,
```
FlorianRappl: We could replace the fancy quotes with standard double quotes - reads the same, but is ready for copy / paste.

text/0027-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+
+All servers will have a config file `config.json` in the `frontends` directory
+of the application data directory. This file can be written to by the server,
+and managed through the UI.
+
+## Definition
+
+An **Extension** is a component with a unique *name* and a *type*.
+
+An **Extension Point** is a place in the DOM where extensions can be attached.
```
FlorianRappl: Why not call it extension slot?

The reason for slot is two-fold. One is, of course, that this is as known from other frameworks (e.g., Piral) the other is that the slot / signal pattern was also made popular in other contexts, which may be helpful for finding the right analogy).

In the end its just a name so it should not matter much - just curious for the reasoning behind point over slot.

text/0027-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+  };
+}
+```
+
+An Extension is an object with lifecycle methods.
+
+Extension Points can be registered in React using the component `ExtensionPoint`
+(and its optional child, `Extension`)
+
+```javascript
```
FlorianRappl: jsx?

text/0027-extensions.md
```diff
@@ -0,0 +1,299 @@
...
+<ExtensionPoint name="foo" type="bar" />
+```
+
+and
+
+```javascript
+<ExtensionPoint name="foo" type="bar">
+  <Extension>
```
FlorianRappl: Potentially `<Extension />`, right?

FlorianRappl (approved):
LGTM - great proposal and work @brandones ! :beers:

text/0027-extensions.md
brandones: Yeah, honestly I kind of prefer "slot" as a name, since it seems to convey more "space" in the UI than "point." The only reason I'm suggesting "point" is because that's the terminology used in the OpenMRS RefApp extension system. If people think that maybe doesn't matter, e.g. maybe it would be good to create some conceptual distance from the old extension system, I would be more than happy to be convinced away from it.

text/0027-extensions.md
brandones: Whoops! That's what I get for starting my work in a Google Doc. Thanks

text/0027-extensions.md
brandones: yes!

text/0027-extensions.md
brandones: Necessarily `<Extension />`, yes. Thanks!

text/0027-extensions.md
brandones: @jonathandick @rbuisson @ibacher Do you have thoughts about the naming?

ibacher: I'm pretty agnostic on this. "Extension point" has some history in the OpenMRS community, not just in the AppFramework, but it's also the name for the equivalent concept in the back-end module implementation. That said, "extension slot" does invoke the idea of a "space" rather than just, e.g., a join point.

jonathandick: slot sounds good to me. would not try to match ref app which already seems to be understood by view

ibacher: I should mention that the backend "extension points" are not, AFAIK, widely used.

jonathandick (approved):

denniskigen (approved):
Very sound proposal @brandones. Excited to work on this with you!

text/0027-extensions.md
```diff
@@ -0,0 +1,307 @@
...
+```jsx
+<ExtensionPoint name="navbarSettings" type="setting" />
+
+// or to add custom render logic
+
+<ExtensionPoint name="navbarSettings" type="setting" />
+  <div className="setting-box">
+    <Extension />
+  </div>
+}/>
```
denniskigen: `</ExtensionPoint>` rounds this bit off nicely

brandones:
```suggestion
</ExtensionPoint>
```

Good catch, thanks!

text/0027-extensions.md
```diff
@@ -0,0 +1,317 @@
...
+```jsx
+<ExtensionSlot name="navbarSettings" type="setting" />
+
+// or to add custom render logic
+
+<ExtensionSlot name="navbarSettings" type="setting" />
+  <div className="setting-box">
+    <Extension />
+  </div>
+</ExtensionPoint>
```
denniskigen: `</ExtensionSlot>`, sorry @brandones

brandones: Thanks!

FlorianRappl:
Can this be merged? Are the 3 votes sufficient or should we wait for Thursday to get more votes + merge?

cc @brandones

text/0027-extensions.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Extension Manager
+
+The extension manager doesn't necessarily have to have its own module, but
+it ought to.
+
+The `<ExtensionSlot />` component
+[should be a](https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent)
+`React.Component`. As in the 
+[current implementation](https://github.com/openmrs/openmrs-esm-api/blob/master/src/shared-api-objects/extension-slot-react.component.tsx#L22),
```
ibacher:
```suggestion
[current implementation](https://github.com/openmrs/openmrs-esm-api/blob/4132f1b25b2044c86fe9fe7ea3c876c9abe214fe/src/shared-api-objects/extension-slot-react.component.tsx#L22),
```

brandones:
@FlorianRappl My thought was that this could serve as a living document until its implementation is complete. I doubt I got everything exactly right—as we learn and improve things in the course of implementing this, we can update the PR. Then when we've implemented the things in this PR we can commit it and have a nice explanation of what we did. We're already moving full steam on implementation: https://issues.openmrs.org/browse/MF-294

FlorianRappl:
@brandones 

Alright I would use this document then also to re-iterate on the activation distribution approach.

I'd deprecate the `lifecycle, activate` exports and include `pages` (as an analogy to `extensions`).

We would have:

```ts
export function setupOpenMRS() {
  return {
    extensions: [...],
    pages: [
      {
        route: "login",
        load: () => import('./my-component'),
      },
    ],
  };
}
```

This offers multiple advantages imho.

brandones:
@FlorianRappl  Just a change to the API, where "lifecycle" becomes "load" and "activate" becomes "route", and now they're wrapped in an array so an MFE can provide multiple apps/pages? Or is this a deeper change? If it's just a simple API change, sure let's throw it in here. If it's something deeper, perhaps it should have its own RFC, just because this RFC is already a bit of a beast. Either way, I'm interested in hearing about the multiple advantages.

Another topic that we should expand on is the packaging distribution stuff you had been talking about.

All of these are part of "activation distribution," but it's a very big theme :)

FlorianRappl:
Thanks for coming back quickly on this one @brandones 

> Just a change to the API, where "lifecycle" becomes "load" and "activate" becomes "route", and now they're wrapped in an array so an MFE can provide multiple apps/pages?

Yes indeed - its just that change. For compatibility reasons we can also leave both `pages` and `activate` / `lifecycle` available for the moment. Long term I would only see this as `pages` (as per the former comment, it provides more flexibility and looks nicer / more familiar when used with `extensions`).

> All of these are part of "activation distribution," but it's a very big theme :)

Fully agreed!

text/0027-extensions.md
```diff
@@ -0,0 +1,320 @@
...
+2. configurability—that implementers can use config files to change which
+    components are plugged in where
+
+We'll build an UI in a new module `@openmrs/esm-implementer-tools`
+for managing that configuration, which will assist implementers
+in changing the config to add, remove, or reorder extensions.
+
+All servers will have a config file `config.json` in the `frontends` directory
+of the application data directory. This file can be written to by the server,
+and managed through the UI.
```
brandones: @FlorianRappl I was going to have module-spa handle this by providing a REST endpoint for writing that file. How do you think we should handle it in the Docker deployment paradigm?

Existing ticket: https://issues.openmrs.org/browse/MF-310

FlorianRappl: Either via a simple Node.js service (a la import map deployer) or just as a static file. In the end the management of this file should not be our concern.

If we would really love to leverage server-side, then this whole project is wrongly set up. However, we fully focus on making many choices on the UI (i.e., on the client) thus a file like `config.json` should be regarded as a static asset. If not, we should not choose `config.json`, but rather some API from the existing backend to use the DB for storing / modifying the config.

brandones: Here is the requirement: an implementer should be able to start an OpenMRS instance with Single-SPA out of the box and start configuring the frontend using the implementer tools and be able to persist those changes in a way that's easy to put in VCS.

I'm going to solve that problem for users of openmrs-module-spa. I will be interested in the approach you take for the others.

brandones:
I'm going to merge this, finally, in the spirit of describing the original plan. The actual system has diverged substantially and should be documented elsewhere.

PR #28, opened 2021-08-18
RFC: Better UX with React-SWR

gracepotma:
Kicking this off with @denniskigen's permission and to continue the conversation we started back in July. 

CC @brandones @FlorianRappl @mks-d @ebambo


FlorianRappl:
I am not sure how using one library gives better UX. It may be that one library helps to create a better UX, but it does not give one.

Also as mentioned right now I don't see any capacity for this. We should first aim to make the solution as bug free and usable as possible. If we would just replace every API call we have with `react-swr` I am also convinced the performance would not be better. We should try to minimize (the number of) requests and make the solution smoother.

Copying my post from Slack:

1. We already use mostly hooks for data fetching (which always reflect the request's life cycle well). Does not help that in most cases we either ignore the error case or simply forward it to the error handler, which only shows a generic message...
2. It does not help with the weight / number of requests that we do right now. Some of these unnecessary requests have already been tamed by proper usage of components (e.g., beforehand we got the patient details like a gazillion times, even though its only required once on the parent component and can then be passed down).
3. Even more conveniently you can declare the resources on a component and the required data will be present when needed. While this will certainly not work in all cases, many cases would benefit from this approach much more (as these resources are then auto-covered for offline mode, too).

Again, the lib is nice and I think we should use it, however, it needs to be understood that this (1) is not a silver bullet, (2) makes only sense if we declare React our primary framework and actively shift in that direction, (3) remove other approaches like using Rxjs.

In any case right now I think we should focus more on stabilizing and making it working instead of refactoring ourselves to death. It's something I'd love to have on a list and tackle, but effort is imho currently spent much better in other areas.

enyachoke-wz:
>  if we declare React our primary framework and actively shift in that direction
I could be wrong about this but I believe one of the advantages  of using MicroFrontends  is to be framework agnostic.

enyachoke:
> if we declare React our primary framework and actively shift in that direction

I could be wrong about this but I believe one of the advantages of using MicroFrontends is to be framework agnostic.

FlorianRappl:
> I could be wrong about this but I believe one of the advantages of using MicroFrontends is to be framework agnostic.

You are wrong about this.

(1) "primary framework" is not "exclusive framework", so don't confuse having a primary framework with not being framework agnostic
(2) You don't need MFs to be framework agnostic - being framework agnostic is a side-effect that is part of many / most / all MF frameworks, but its by no means a necessity (still it's an aspect we'd keep in any case)
(3) Hosting multiple frameworks has severe performance (and maintainability) impact - it makes most sense in open systems as we have one (which is why we should keep this aspect)

The bottom line is quite simple: Since we develop almost all MFs (at least up to now) and since we exclusively use React we should make that part as fast and easy to develop as possible. Right now it's neither, even though the core framework already comes with React-exclusive helpers and React-exclusive functionality. We have a dozens of extensions floating around, but all of them need to spawn a new React render tree, even though a single tree with a single context would be sufficient... Making React the primary framework would give a huge boost here, without putting other frameworks at a disadvantage. They would have the same metrics as now.

text/0028-data-updates.md
```diff
@@ -0,0 +1,78 @@
...
+* How it works: immediately loads cached data, while simultaneously checking for any changes/updates. 
+* Implications: Means user can see updated condition list after they save their change.  
+* *Relevant PR*: This can probably be cleaned up a bit but this should be an accurate 
+    representation of the work done to get use-swr working  https://github.com/openmrs/openmrs-esm-patient-chart/commit/4563425a859b8fd65f5546fccb941afda7546bd8 
+
+### Caveats
+We must understand that this approach:
+1. is not a silver bullet
+2. only makes sense if we declare React our primary framework and actively shift in that direction
+3. requires removing other approaches like using Rxjs.
```
gracepotma: @FlorianRappl this was already added ;)

enyachoke:
> > I could be wrong about this but I believe one of the advantages of using MicroFrontends is to be framework agnostic.
> 
> You are wrong about this.
> 
> (1) "primary framework" is not "exclusive framework", so don't confuse having a primary framework with not being framework agnostic
> (2) You don't need MFs to be framework agnostic - being framework agnostic is a side-effect that is part of many / most / all MF frameworks, but its by no means a necessity (still it's an aspect we'd keep in any case)
> (3) Hosting multiple frameworks has severe performance (and maintainability) impact - it makes most sense in open systems as we have one (which is why we should keep this aspect)
> 
> The bottom line is quite simple: Since we develop almost all MFs (at least up to now) and since we exclusively use React we should make that part as fast and easy to develop as possible. Right now it's neither, even though the core framework already comes with React-exclusive helpers and React-exclusive functionality. We have a dozens of extensions floating around, but all of them need to spawn a new React render tree, even though a single tree with a single context would be sufficient... Making React the primary framework would give a huge boost here, without putting other frameworks at a disadvantage. They would have the same metrics as now.

Understood thanks for clarifying this.

brandones:
> We have a dozens of extensions floating around, but all of them need to spawn a new React render tree, even though a single tree with a single context would be sufficient... Making React the primary framework would give a huge boost here, without putting other frameworks at a disadvantage. They would have the same metrics as now.

This is quite compelling. Why don't you open an RFC for this? And then we can have this RFC be about SWR. As you say,

> the core framework already comes with React-exclusive helpers and React-exclusive functionality

Therefore the question of whether or not to adopt SWR doesn't impact our relationship to React in a new way. It's just more of what's already been happening.

FlorianRappl:
> Therefore the question of whether or not to adopt SWR doesn't impact our relationship to React in a new way. It's just more of what's already been happening.

In this case there is no need for the RFC. Nothing prevents anyone from just using a library in any MF. But as we discussed (feel free to correct me if I don't recall this correctly) this is not what we want. We want to have a single consistent and reusable way shared by pretty much all (React) MFs.

brandones:
> We want to have a single consistent and reusable way shared by pretty much all (React) MFs.

Exactly. So let's discuss whether we should adopt SWR as that consistent and reusable way.

brandones:
@jonathandick @mks-d @denniskigen ?

gracepotma:
Another example for why we need to address this performance problem: I closed https://issues.openmrs.org/browse/MF-714 because the problem doesn't exist after refreshing the page (the drug information is updated). However, **in the real world, our clinical users will not go to the trouble of refreshing their pages!!** 

We need to move a solution forwards. It is not clinically acceptable for the UI/UX to seem out of date to the user (e.g. if a change is made in the workspace; we really do need the relevant widget in the body to update near-instantly).

FlorianRappl:
@gracepotma what you describe is not a performance problem, but a UX problem. Data views that can get stalled should either auto-update after a while (if they are still active / used) or have a refresh button close to them (which would only refresh the data in question, and not reload the whole app).

> However, in the real world, our clinical users will not go to the trouble of refreshing their pages!!

I am not sure on that. This does not make much sense as "classically" the only way to get updated information on a page is to refresh it. Naturally, there is even a special gesture on smartphone to refresh a page (guess why).

Note that this has nothing to do with the frontend. Even when we ping the backend in periodic time ranges we will do a wrong thing. Let's say we ping every 10 seconds - well if an update was made after 1 second you will be 9 seconds late. In most cases there is no update - which results in an unnecessary request.

To support this case, backends usually provide an event channel; either via server-sent events or websockets. I am not aware that the OpenMRS backend supports either one - but this would be required to have the best UX (and the one you described / apparently desire).

brandones:
No, the thing that @gracepotma is asking for is [exactly what the SWR library does](https://swr.vercel.app/docs/mutation). Re-read what she wrote:

> if a change is made in the workspace; we really do need the relevant widget in the body to update near-instantly

This does not require new backend support. And it is a much better UX than anything you are proposing.

It seems like a lot of your comment is addressing the problem of how to get changes to a patient's chart to sync *between clients* in real time. This is out of scope for the present discussion; as you say, backend changes would be needed. Feel free to open a different RFC if you want to talk about that problem.

FlorianRappl:
My comment is addressing the sync of data, not the sync of clients. The sync of data is the real problem here. What you describe is nothing that cannot be handled already - generically, and independent of the method of transportation.

Before we go in circles again I'd like to remind you that there is nothing against using React SWR already in micro frontends (as mentioned above). I am not sure why we still discuss rolling it out on the framework level, since no other cleanup (e.g., removing of Rxjs) has yet been started.

So let's close or pause the RFC and have somebody to roll out React SWR in the patient chart monorepo.

brandones:
> So let's close or pause the RFC and have somebody to roll out React SWR in the patient chart monorepo.

Good enough for me. CC @denniskigen

brandones:
Rewrote for clarity, to remove some language that was false ("requires removing RxJS") or otherwise meaningless and denigratory ("not a silver bullet"), and to add a handful of alternatives that we more or less considered.

Let's merge with **one approving review**, to document what has already in practice been adopted.

denniskigen (approved):
Thanks a bunch for clarifying this information, @brandones.

PR #29, opened 2021-08-18
RFC: Common Widgets Directory

gracepotma:
Kicking off discussion - we should just try this out!!

Kudos to @ebambo for this idea.


FlorianRappl:
Hm I do not really understand the proposal. Maybe you can clarify what a widget here (is it an extension, or any component)?

PR #30, opened 2021-10-22
RFC: Link system

brandones:


text/0030-link-system.md
```diff
@@ -0,0 +1,136 @@
...
+framework-independent functions that would support it). `Menu` would
+work exactly like `ExtensionSlot`, but only for Links.
+
+```typescript
+function Menu(props: { name: string, state?: object });
+```
+
+We will extend the configuration system so that every Menu supports
+the same configuration options as ExtensionSlots, but without `configure`, and
+with the addition of a new key, `create links`:
```
ibacher: It might be simpler to just name this key `links`? I think that also reads better from the perspective of "I'm an implementer and I want to add links to this menu".

text/0030-link-system.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+For links created in this way, the link label is the ID.
+
+## Reason for decision
+
+- We have a proliferation of trivial "link" extensions, which all share
+  approximately the same simple structure.
+- Link styling should be left up to the slot/menu. This enforces that
+  links are not bringing their own `class`es.
```
ibacher:
```suggestion
  links are not bringing their own CSS classes.
```

text/0030-link-system.md
```diff
@@ -0,0 +1,136 @@
...
+}): () => Promise<ReactAppOrParcel<any>>
+```
+
+Where `to` can include `${variables}` which will be interpolated from
+`openmrsBase`, `openmrsSpaBase`, or the extension props; and where
+`onClick` receives extension props as its second argument.
+
+This can be used in an extension definition, for example:
+
+```typescript
```
ibacher:
```suggestion
```ts
```

text/0030-link-system.md
```diff
@@ -0,0 +1,136 @@
...
+      })
+    }]
+  }
+}
+```
+
+In cases where the extension `meta` property is not required, the
+following could equivalently be used:
+
+```typescript
```
ibacher:
```suggestion
```ts
```

text/0030-link-system.md
```diff
@@ -0,0 +1,136 @@
...
+    }]
+  }
+}
+```
+
+We will define a new React component, `Menu` (along with
+framework-independent functions that would support it). `Menu` would
+work exactly like `ExtensionSlot`, but only for Links.
+
+```typescript
```
ibacher:
```suggestion
```ts
```

ibacher (commented):
Should we say something about how the single link extension should work? I.e., as I see it there are likely three different kinds of links implementers might want:

1. Links to another part of the SPA
2. Links to a part of the legacy UI
3. Links to an external service

I assume that single-spa's routing can handle 1 and that links to one of the legacy UIs work as-is, but for external links, we probably want the link generated with the `target="_blank" rel="noopener noreferrer"` attributes

text/0030-link-system.md
brandones: My fear is that implementers would think "oh, this is the canonical list of all the things that go in here, which are links." But actually I think you might be right, that your way of reading it is probably more organic.

brandones:
Good thought, @ibacher . Should there be a property `openInNewTab: boolean` which does `target="_blank"`, so that implementers can customize which links open in a new tab (default true for external links)? Or is just doing it for external links pretty sure-fire? I'm thinking of thinks like where PIH Mexico has some informational PDFs that serve from the OpenMRS server that they want to open in a new tab when the link is clicked.

What's annoying is that this would seem to suggest that we would want to make links "configurable" with this one property, which is something I was hoping to avoid, but maybe is inevitable.

ibacher:
> but maybe is inevitable

Well, the good part is that there's not a lot else you can do with an anchor tag that seems worth supporting here. But, now that you've mentioned it, I do think having this as a configurable property is something we'll need, i.e., it's probably not a great idea to be trying to guess what someone meant.

brandones:
Cool. Should the `rel="noopener noreferrer"` attribute always be added when `openInNewTab == true`, or only for external links?

ibacher:
I think it's fine if it's only for external links... that's really where the risk is.

ibacher (approved):
LGTM

brandones:
@denniskigen @donaldkibet @jonathandick @manuelroemer @samuelmale

text/0030-link-system.md
```diff
@@ -0,0 +1,148 @@
...
+          "Patient list link": {
+            "openInNewTab": true
+          }
+        }
+      }
+    }
+  }
+}
+```
+
```
jonathandick, line 76: @manuelroemer how would this be handled in offline scenario? Should there be a property to specify offline availability?

brandones: It's a good question—should links appear and disappear based on whether the user is offline? Can we think of any examples of links that should do that?

brandones: @jonathandick @mks-d @manuelroemer

jonathandick: My understanding right now is that links _do_ appear and disappear in the patient chart based on whether you are online or offline though I don't know how this is done. @manuelroemer could you comment?

jonathandick (commented):
Two uses cases of concern for me:

First, we (eventually) want to be able dynamically create a page (via configuration) and a link to a menu to get to that page. Will these menus be able to accommodate this? 

Secondly, would this be used to support something like the actions menu within various patient chart widgets? Or is it meant to apply for higher level navigation like the left menu bar or the app switcher?

brandones:
Great questions, thanks @jonathandick !

> First, we (eventually) want to be able dynamically create a page (via configuration) and a link to a menu to get to that page. Will these menus be able to accommodate this?

This wouldn't make it any easier nor harder than it presently is. Whereas without Link Extensions we would be dynamically creating a link-like extension, with Link Extensions we would dynamically create an actual link extension.

> Secondly, would this be used to support something like the actions menu within various patient chart widgets? Or is it meant to apply for higher level navigation like the left menu bar or the app switcher?

Any of the above. If we'd like it to be configurable, this is a way to make that easy.

jonathandick (approved):
Great! Thanks @brandones

denniskigen:
LGTM @brandones!

denniskigen (approved):

text/0030-link-system.md
manuelroemer: @jonathandick Yes, that's true! The patient chart is a good example for a menu where some links aren't available when offline. This behavior must be possible.
I want to remark that extensions *are not* available offline unless they opt into offline mode (i.e. `offline` is `false` by default unless specifically configured otherwise). This means that we require the opposite behavior: We want links to be able to specifiy that they *should* render in offline mode.

However, given that the links are still normal extensions (albeit abstracted away), it should already be possible to specify the `online`/`offline` configurations with the registration proposed by @brandones:
```diff
function setupOpenMRS() {
  return {
    pages: [],
    extensions: [{
      id: 'Patient list link',
      slot: 'App menu',
      load: createLinkExtension({
        label: () => t("patientList", "Patient List"),
        to: "${openmrsSpaBase}/patient-list/${patientUuid}",
        onClick: (e, { patientUuid }) => {
          console.log(`Clicked ${e.button} for patient ${patientUuid}`);
        }
      }),
+     online: true,
+     offline: true
    }]
  }
}
```

manuelroemer:
Hey, sorry for the late comment - not sure if it's even necessary anymore since 3 approvals are already there. I think that the proposal is good in general and will drastically reduce boilerplate code in certain areas. I do have some thoughts and questions after reading through the RFC though and wanted to share them:

#### Link Registration
In the proposal there are two different ways of registering links: (1) as extensions and (2) via the new `links` attribute. In my opinion it might be a good idea to just focus on one of those in the beginning, preferably (1). My reasoning is that less options are less confusing and that it will allow registered links to easily leverage all features of the extension API. The `online`/`offline` attributes are good examples: Since the links are just plain extensions in the end, they can use these configurations. In my opinion it makes little sense to mirror these into a new `links` configuration when it already works perfectly fine with the `extensions`.

#### The Menu
Some questions:
To me it's not entirely clear how the `Menu` component will work. I assume that I can simply drop it anywhere, give it a unique name and that it will automatically render every link registered for it. Is that correct?

> Link styling should be left up to the slot/menu. This enforces that links are not bringing their own CSS classes.

Is it possible to hook into the menu or change how it renders the links (e.g. make them smaller if desired)? Is this something that is required or could become a requirement in the future?

brandones:
Thanks so much @manuelroemer . I was hoping to be able to integrate your feedback, especially around offline stuff, before merging. This is really helpful.

> In the proposal there are two different ways of registering links: (1) as extensions and (2) via the new links attribute. In my opinion it might be a good idea to just focus on one of those in the beginning, preferably (1). My reasoning is that less options are less confusing and that it will allow registered links to easily leverage all features of the extension API. The online/offline attributes are good examples: Since the links are just plain extensions in the end, they can use these configurations. In my opinion it makes little sense to mirror these into a new links configuration when it already works perfectly fine with the extensions.

That's very fair. I agree that we should just pick one, since that will be less confusing. I think my inclination, though, is to use the `links` attribute rather than the extension-making function. This is mostly because "Menu" is its own piece of the API, distinct from "ExtensionSlot." So we would have this nice correspondence "extensions --> Extension Slots" and "links --> Menus." Even if under the hood, it's all extensions and extension slots. And technically the implementation is very similar between (1) and (2)—the `links` attribute is just a thin wrapper over the extension registration API. So "links" will already be able to leverage all the features of the extension API.

I can see a good argument for doing it either way. @ibacher @jonathandick @denniskigen ?

> To me it's not entirely clear how the Menu component will work. I assume that I can simply drop it anywhere, give it a unique name and that it will automatically render every link registered for it. Is that correct?

Yeah, exactly. Just like an Extension Slot.

> Is it possible to hook into the menu or change how it renders the links (e.g. make them smaller if desired)? Is this something that is required or could become a requirement in the future?

I'm not sure I understand exactly what you mean. I think that generally, the menu should control the style, because the same link in different menus should be styled according to which menu it is in.

But if you mean that it should be possible for links to differentiate themselves within menus, yeah, I think that is a problem we need to be able to solve, but which isn't solved in this RFC. @jonathandick told me about a use case with OHRI where they have links contained in accordion headers:

image

My suggestion for how to solve this is to include that kind of information in the link's `meta`. So say `meta.header`. And then I think we should define a `render` parameter on `ExtensionSlot`s and `Menu`s that would allow you to provide a custom rendering for the extensions that uses their `meta`. Something like

```jsx
<Menu name="Patient chart left nav" render={links => {
  const categories = new Set(links.map(l => l.meta.category));
  return categories.map(c => <>
    { links
        .filter(l => l.meta.category == c)
        .map(l => l.render()) }
    </>)}
/>
```

This is of course inspired by the way Piral handles extensions.

ibacher:
> I agree that we should just pick one, since that will be less confusing.

I agree with the picking one part; I tend to agree with @brandones that having a distinction between "slots" and "menus" is preferable, but this is a weak preference and I can certainly see arguments to stick with just the normal extension configuration.

brandones:
Right, so I'm remembering the reason I made two mechanisms in the first place: the "extensions way" of defining a link was meant to be more semantically sensible when doing dynamic page creation. That is, we have this kind of strange pattern where we define extensions which are themselves literally just links, but which contain `meta` properties that a module uses to generate a page (the page that the link points to).

I think this is a pretty weird use-case, and the right answer is probably actually to design a better abstraction for it specifically, which is neither a standard `extension` nor a `link`.

For now, I'm going to remove the function for making a link in an extension definition.

Thank you all very much for helping figure all this out!

PR #33, opened 2026-08-04
RFC: Static configuration schemas

ibacher:
This proposes extracting configuration schemas as JSON at build time and combining them into the existing `routes.registry.json`, allowing these schemas to be available in the application without loading any JS.


text/0032-static-config-schemas.md
```diff
@@ -0,0 +1,42 @@
+# Static configuration schemas
+- Start Date: 2026/08/04
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/29
+
+## Decision, including impact on distributions
+
+Frontend module configuration schemas are static data, not the result of executing module code:
+
+- Each module publishes its config schema as a build artifact (`config-schema.json`, alongside `routes.json`). `openmrs assemble` merges each module's schema into its entry in `routes.registry.json` — authored separately, delivered as part of the single registry the shell already fetches at boot — and the configuration system initializes from it before any module loads.
+- Loading a module is no longer a precondition for resolving its configuration, including extension slot configuration provided by that module.
```
denniskigen, line 12: Can we include extension config schemas in the artifact contract too? `esm-config` [resolves them by extension name](https://github.com/openmrs/openmrs-esm-core/blob/3757bfe51779418c5e407d7b5ee4097a8104d1bb/packages/framework/esm-config/src/module-config/module-config.ts#L428-L439), and apps can register several schemas from files other than `src/config-schema.ts` ([example](https://github.com/openmrs/openmrs-esm-patient-chart/blob/e95dcece7edf42db2315038e4efd2a4ff24e5d3d/packages/esm-generic-patient-widgets-app/src/index.ts#L20-L22)). With only the module schema described here, those defaults and descriptions still require `startupApp` to run. It's probably best to define how they're discovered and keyed in `config-schema.json`.

text/0032-static-config-schemas.md
```diff
@@ -0,0 +1,42 @@
+# Static configuration schemas
+- Start Date: 2026/08/04
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/29
```
denniskigen:
```suggestion
- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/33
```
This currently links to #29, the Common Widgets Directory RFC.

text/0032-static-config-schemas.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Decision, including impact on distributions
+
+Frontend module configuration schemas are static data, not the result of executing module code:
+
+- Each module publishes its config schema as a build artifact (`config-schema.json`, alongside `routes.json`). `openmrs assemble` merges each module's schema into its entry in `routes.registry.json` — authored separately, delivered as part of the single registry the shell already fetches at boot — and the configuration system initializes from it before any module loads.
+- Loading a module is no longer a precondition for resolving its configuration, including extension slot configuration provided by that module.
+- Schema validators are expressed as references to a built-in vocabulary, which is expanded to cover the common cases (`greaterThan`, `positiveInteger`, `nonEmptyString`, `minArrayLength`, joining the existing built-ins).
+- Custom cross-field validators remain ordinary functions, in a conventional module (`config-validators.ts`) that the framework loads off the boot path — after first render, in the implementer tools, or at assemble time. Implementer-facing validation output is unchanged.
```
denniskigen: Can we spell out the export and discovery contract for this module? Today the association between a validator and its schema path is encoded directly in that node's [`_validators` array](https://github.com/openmrs/openmrs-esm-core/blob/3757bfe51779418c5e407d7b5ee4097a8104d1bb/packages/framework/esm-config/src/module-config/module-config.ts#L739-L759). Since the functions can't ship in `config-schema.json`, we need a mapping that lets the framework reattach and rerun them. Otherwise, the unchanged validation output isn't specified.

text/0032-static-config-schemas.md
ibacher: This is a very good point and something I hadn't really considered. I think my first pass at an answer here is that apart from being a little more careful in wording, this is probably a non-issue for the RFC as such. This RFC doesn't actually specify the _content_ of `config-schema.json`. It wouldn't necessarily be hard to make this something like:

```json
{
	"module": {
		// config schema goes here
	},
	"extensions": {
		"extension1": {
			// config schema goes here
		}
	}
}
```

What this _does_ bring up, however, is how we deal with the global namespacing of extensions and, specifically, what happens if two apps / microfrontends define an extension schema with the same name. Right now, the extension config belongs to whichever `defineExtensionConfigSchema()` happened to execute first. With the shift to JSON we could. My instinct here would be to add a phase to assemble that would catch these issues and warn loudly about them, but roughly still keep the first-loaded as the winner.

---

A related, but distinct point is that the current arrangement makes it _possible_ (though annoying) to define an _arbitrary_ configuration schema associated with an _arbitrary_ module key. I'm not aware of anyone actually doing that, but should we consider the possibility that one "app" could have multiple configuration schemas it defines or should we rule that out here?

text/0032-static-config-schemas.md
ibacher: I think this was actually in some of the implementation details I stripped out for this version of the RFC. That said, one thing to remember is that `config-schema.json` _ultimately_ ends up in `routes.registry.json` and, just like `routes.json`, from the framework's perspective that's ultimately the only place that matters. That means that what actually gets consumed at runtime looks more like this:

```json
{
	"@openmrs/esm-arbitrary-app": {
		"backendDependencies": {},
		"extensions": [],
		"configurations": { "_comment": "content from config-schema.json goes here" }
	}
}
```

So, while we may not have a direct reference to the schema, internally, we can re-write the validators as something like `@openmrs/esm-arbitrary-app/custom-validators#customValidator1`, which gives us a way of locating that validator unambiguously (and note that validator implementations end up in their own, separately addressable "chunk").

ibacher: One point I didn't raise, but maybe is worth discussing is that our current validators are actually toothless. All we do with them is log an error message to the console if they fail (the only place I could find that we directly use the return value  is [here](https://github.com/openmrs/openmrs-esm-core/blob/16a6c01d099267045f174f95ecf3936d3b770fb9/packages/framework/esm-config/src/module-config/module-config.ts#L717), where we just use it to skip further processing something that doesn't look like an array.

This feeds into the RFC only in as much as I was contemplating moving the _timing_ of validating so that it's done at a later point, either before a configuration is loaded for the first time or maybe when someone launches the implementer tools or even as part of the [implementer tools validation](https://github.com/openmrs/openmrs-esm-core/blob/16a6c01d099267045f174f95ecf3936d3b770fb9/packages/apps/esm-implementer-tools-app/src/configuration/interactive-editor/validators.resource.ts#L125).

chibongho (commented):
Thanks. I like with the overall idea. 

Somewhat tangential point copied from the #o3-leads discussion, it'd be nice if we can have:
- a better way to share configs across different modules, perhaps via a global config scope:
  - `drugOrderTypeUUID` is duplicated in esm-patient-medications-app and esm-dispensing-app
  - `customPatientChartUrl` is duplicated in esm-appointments-app and esm-service-queues-app
  - Things like `dateFormat` (MM/DD/YYYY) would be a good candidate for global scope config

  Moving the config schema to routes.json should make it easier to guarantee that the configs are available at runtime (without having to load the app first), so across-app config sharing should be easier.
- more `Types` available in the config schema, like:
  - Type.Icons
  - Type.Color

ibacher:
@denniskigen I've updated the text to address your concerns. Could you let me know if that's sufficient. One thing I'm deferring for a different RFC is extension configuration collisions. I think these are solvable in a principled way via the assemble configuration, but the same solution should be applied to extension _definition_ conflicts as well, so I'm deferring it to another RFC.

@chibongho Agreed on all of this. I think shared configs is something we should solicit input on (is the solution cross-module config imports? "Special" namespace à la `useConfig('shared:medications')`? A genuinely "global" configuration whatever that means in context? Should we push all these things into Global Properties?). I.e., I think this will become a new RFC draft.

On the types, I think we have a ticket for that already? That should be enough. On the `Type.Icon`, I've disgusted with Dennis (but not implemented) leveraging Claude to do the re-writing so we can completely move all icon loading into the framework (basically to try to save ourselves bytes). That would also make it easier to generate, e.g., the valid values for such a `Type`. 

For `Type.Color`, I can think of three "types" of use cases: one for Carbon Tags where we're dealing with a [very limited vocabulary](https://github.com/carbon-design-system/carbon/blob/476382bee047ff199c01c9939cda703c2eadc551/packages/react/src/components/Tag/Tag.tsx#L33) of possible colors, a more generic "string used in a CSS property" and maybe a more limited "a few CSS-like options" (like six-character hex-strings + all named colors). What direction were you thinking of? Or should it be "any of those with a validator to discriminate if the input is right?". FWIW, Carbon Tags are, I think, a strict subset of CSS named colors, so that's the most restrictive sense.

ibacher:
For posterity, I also want to note that Joseph Kagimu (@Josephkagimu1) also [commented on talk](https://talk.openmrs.org/t/o3-rfc-static-configuration-schemas/49895/2) asking:

> One question: during the transition window where defineConfigSchema still works, will modules with existing custom validators need any changes, or will the build extraction handle those automatically until we migrate to config-validators.ts?

Two points I made in answer:

* `config-validators.ts` is mostly a suggestion about code structure; nothing should depend on validators being located there (though, I think as-written the implied _implementation_ will be creating something like a "config-validators" chunk, but that's a different matter)
* We will do our best to support _all existing code_ without changes here, though there are potentially some edge-cases where Webpack / RSpack wouldn't be able to correctly identify all validators.

On that latter point, it's worth noting that the same caveat applies to existing config schemas. We don't want to build machinery that will assume the config schema is completely located in `config-schema.ts` or the like, but there are theoretical cases (like where the configuration schema depends on the result of `eval()`-ing an expression whose value cannot be statically determined) where we may not be able to do this automatically. No such cases exist in the O3 RefApp nor in a few community modules I examined, but I'm inclined to say that those will be handled by failing the build loudly.

ibacher:
Last point: there is always a manual escape hatch. Ultimately, this will wind-up as another part of the `routes.registry.json` so it's possible for a module that doesn't like our implementation to:

* Override it it by generating the corresponding `dist/config-schema.json` at any point before publishing their app (or converting it into a `.tgz`. As long as the `config-schema.json` is something the framework can parse and understand, it will honor it.
* Doing the overrides by modifying `routes.registry.json` which will technically work but I want to discourage in favour of the above suggestion.

I should also say: we will publish a JSON schema for `config-schema.json` just as we do for forms and routes.json. We will continue not to publish JSON schemas for `routes.registry.json` as this file should be treated as an opaque descriptor.

chibongho:
> For `Type.Color`, I can think of three "types" of use cases:

I was thinking Carbon Tag colors. I have a use case for the ward app to display particular obs as tags in the patient card.

image

That said, I'm not opposed to having another type for more genetic CSS colors too.

denniskigen:
Thanks, @ibacher. The extension schema keying and the single-module-schema rule address my concerns. Two small things:

- Can we state which schema wins when an extension name collides and the build isn't strict? Your comment says first-loaded wins - it'd be good to have that in the text so the behavior is specified even with full conflict resolution deferred to the future RFC.
- Can we fold the validator reference form (`@openmrs/esm-arbitrary-app/custom-validators#customValidator1`, sitting where the function sat) into the text? That's the detail that makes "implementer-facing validation output is unchanged" concrete.

text/0032-static-config-schemas.md
denniskigen: Ruling out arbitrary module keys in the artifact seems right - the runtime API covering it during the transition window is a good escape hatch.

denniskigen (approved):

ibacher:
> Can we fold the validator reference form into the text?

I'd rather not commit us to that specific syntax, but I've added some wording to capture the idea.

> Can we state which schema wins when an extension name collides and the build isn't strict?

I've added a sentence specifying this. The order is, ultimately, the order things appear in `routes.registry.json`.

PR #34, opened 2026-08-12
RFC: Extract CSS from JS

ibacher:
This RFC essentially requests that we remove `style-loader` from production code while continue using it in development mode. This largely means that the CSS code "lands" in a CSS file rather than CSS-in-JS.

### DIFF
--- a/text/0033-extracted-css.md
+++ b/text/0033-extracted-css.md
@@ -1,0 +1,35 @@
+# Extracted CSS
+- Start Date: 2026/07/30
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/TBD
+
+## Decision, including impact on distributions
+
+Production builds of frontend modules emit CSS as real `.css` assets instead of embedding it in JavaScript:
+
+- The shared build configs use CSS extraction in production and keep `style-loader` in development.
+- Carbon's CSS is delivered exactly once, by the app shell. Frontend module builds must not emit Carbon rule definitions; the build enforces this by checking emitted CSS.
+
+Modules adopt this with a tooling bump — no source changes, no visual changes. Distributions require no changes.
+
+## Definition
+
+Today, `style-loader` embeds each module's compiled CSS as strings inside JavaScript chunks and injects `<style>` tags at runtime. Extraction instead emits per-chunk `.css` files that the module's own runtime loads alongside its JavaScript chunks; a chunk does not resolve until its CSS has loaded, so components never mount unstyled. CSS Modules class-name generation is unchanged — only delivery changes.
+
+## Reason for decision
+
+- Frontend modules currently ship 300–800 kB of CSS *as JavaScript* per app (measured on the reference application), paying JS download, parse, and execution costs for what is stylesheet content. Extracted CSS parses on the browser's CSS path and caches independently of code changes.
+- Real CSS assets can be preloaded per route; CSS trapped in JS cannot.
+- Measurement shows no module currently duplicates Carbon's rules (app CSS only *references* Carbon classes in overrides), so the Carbon guard codifies the existing status quo rather than forcing migrations.
+- Development keeps `style-loader` because dev tooling is optimized for developer experience; it is the framework's responsibility that the default configurations behave equivalently across modes.
+
+## Alternatives
+
+- **rspack's native CSS support** — likely the long-term direction, but the webpack and rspack shared configs are kept behaviorally identical while webpack compatibility is retained.
+- **Extracting in development too** — rejected; worse HMR experience for no user-facing benefit.
+- **Keeping style-loader and layering critical-CSS techniques on top** — treats the symptom; CSS still ships and parses as JS.
+- **Linting Carbon imports at the source level** — brittle against the many valid token/mixin import paths; checking emitted output verifies the thing that actually matters.
+
+## Common practices (not enforced)
+
+- Module SCSS should `@use` only Carbon tokens, mixins, and functions (which emit no CSS), never Carbon component styles or `@carbon/styles` wholesale.
+- Modules that override the shared configs' style rules own the consequences; the defaults are the supported path.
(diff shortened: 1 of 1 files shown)


denniskigen (commented):
This is in good shape, @ibacher. I prototyped the extraction path on two apps in patient-chart and left a few notes on the measurements and rollout assumptions.

text/0033-extracted-css.md
```diff
@@ -0,0 +1,35 @@
...
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/34
+
+## Decision, including impact on distributions
+
+Production builds of frontend modules emit CSS as real `.css` assets instead of embedding it in JavaScript:
+
+- The shared build configs use CSS extraction in production and keep `style-loader` in development.
+- Carbon's CSS is delivered exactly once, by the app shell. Frontend module builds must not emit Carbon rule definitions; the build enforces this by checking emitted CSS.
+
+Modules adopt this with a tooling bump — no source changes, no visual changes. Distributions require no changes.
```
denniskigen: Shouldn't we qualify this for distributions that serve frontend modules from a different origin? `publicUrl` and absolute import-map entries support that setup. Today a deployment can allow that origin only in `script-src` while permitting `style-loader`'s injected `<style>` via `'unsafe-inline'`. Extracted `<link>` stylesheets use `style-src-elem`, falling back to `style-src` and then `default-src`, so the affected chunk fails to resolve unless that origin is allowed there too. It's probably best to call out the CSP update required for off-origin module hosting.

text/0033-extracted-css.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Definition
+
+Today, `style-loader` embeds each module's compiled CSS as strings inside JavaScript chunks and injects `<style>` tags at runtime. Extraction instead emits per-chunk `.css` files that the module's own runtime loads alongside its JavaScript chunks; a chunk does not resolve until its CSS has loaded, so components never mount unstyled. CSS Modules class-name generation is unchanged — only delivery changes.
+
+## Reason for decision
+
+- Frontend modules currently ship 300–800 kB of CSS *as JavaScript* per app (measured on the reference application), paying JS download, parse, and execution costs for what is stylesheet content. Extracted CSS parses on the browser's CSS path and caches independently of code changes.
```
denniskigen, line 20: I tried this on a clean patient-chart checkout by replacing `style-loader` with `CssExtractRspackPlugin`. Both builds completed without source changes. The other CSS assets totalled about 9 kB for flags and 97 kB for patient-chart, but both builds also emitted an 894 kB stylesheet for the local `@openmrs/esm-framework` fallback.

That asset is fetched only if Module Federation selects the fallback. Was the 300–800 kB measurement network-observed, or does it include packaged fallback CSS? Can we make that distinction here and track fallback hygiene separately?

text/0033-extracted-css.md
```diff
@@ -0,0 +1,35 @@
...
+- Real CSS assets can be preloaded per route; CSS trapped in JS cannot.
+- Measurement shows no module currently duplicates Carbon's rules (app CSS only *references* Carbon classes in overrides), so the Carbon guard codifies the existing status quo rather than forcing migrations.
```
denniskigen, line 22: Can we update this? I found five emitting imports in two reference-app modules: three `@carbon/styles/scss/components/button` imports in `esm-implementer-tools-app`, plus `_flexbox` and `@carbon/charts/styles.css` in `esm-patient-tests-app`. The charts CSS is already shipped globally by the styleguide.

The shared CSS Modules config hashes these class selectors, so they do not collide with the shell's global Carbon rules, but the generated payload is still emitted. A guard anchored on `.cds--` at a selector boundary would miss these copies. If scoped copies count, these five imports require source changes, so this bullet and the "no source changes" claim both need updating.

text/0033-extracted-css.md
ibacher: Yes. The claim here isn't that we're saving 300-800 kB of data. It's that there are 300-800 kB of CSS-in-JS per module, but you're right that this is dominated by the styleguide. Realistically, though, this change isn't designed to save bytes; at most, this would only ever save us a handful. The goal is to improve performance by passing-off to the CSS to the browser's CSS parser without the JS parser in the mix.

text/0033-extracted-css.md
ibacher: The idea isn't that we aren't re-using Carbon SCSS files anywhere; it's that we're not redefining them anywhere (other than, the Carbon Charts thing since that's separately vendored per app). So, e.g., in the implementer-tools-app, we're not redefining `.cds--button` or related classes; we're just using the _color_ token from the button SCSS.

ibacher: The thing this is trying to avoid is that if we were overriding Carbon's CSS classes, then we'd potentially have issues with needing to unload CSS when a module unloads, but if we don't (i.e., the CSS selectors created by the built code are not effecting the global Carbon styles or only do so with app-specific selectors) then there's less to worry about.

text/0033-extracted-css.md
ibacher: Yes, that's a good point.

ibacher: A mitigation here is that the _app shell_ already loads it's CSS as separate files, so it would only affect the situation where:

1. There's a restrictive CSP
2. Frontend components are loaded from more than one origin.

denniskigen (approved):
LGTM

denniskigen:
One implementation follow-up: I rechecked the current Carbon Sass entrypoints with CSS extraction enabled. The button and flexbox entrypoints emit Carbon rules, and the patient-tests Charts import emits another copy. CSS Modules hashes these selectors, so they do not collide with the shell’s global rules, but the rule definitions and payload are still duplicated.

This means the emitted-CSS guard needs to recognize hashed Carbon selectors, and the affected imports will need cleanup rather than only a tooling bump. Could we track this alongside the implementation work?

PR #35, opened 2026-09-04
0034-extension-preconditions

ibacher:
Frequently, we have extensions that only need to render if certain data are available. An example is the deceased tag, which only renders anything if the patient is actually deceased. As things stand, however, these extensions still have to be mounted in order to make this determination and this mounting is not strictly free. The proposal here is to provide a mechanism for extensions to tell the framework when they need such things resolved and resolve them _before_ the extension is actually rendered.

Things I'm uncertain about here:

* This could become a standard data pre-fetch mechanism. As proposed, it would basically work that way if the `"when"` condition were `true`. It may be that we want to make this intentional rather than a side-effect. I'm somewhat uncertain how we could make this work _cleanly_ with SWR. The advantage of data pre-fetches would be that we could do de-duplication directly in the framework and not strictly rely on SWR doing so.
* This proposal changes the definition of assigned extensions in a way that adds a potential data-fetch overhead to any use of this. While I don't think it's common to use assigned extensions outside of the framework, it is something we've had to reach for in some advanced cases for, e.g., dashboard extensions on the patient chart. As long as assigned extensions are used in the same render that also renders the ExtensionSlot, I don't think there's a risk of duplicate requests, but usages outside of that need to be considered.

Since RFCs are sometimes hard-to-read, I've also added a [draft ticket](https://openmrs.atlassian.net/browse/O3-5956) that describes slightly more implementation-specific detail, which may be easier to review.

### DIFF
--- a/text/0034-extension-preconditions.md
+++ b/text/0034-extension-preconditions.md
@@ -1,6 +1,6 @@
 # Extension Preconditions
 - Start Date: 2026/09/04
-- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/TBD
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/35
 
 ## Decision, including impact on distributions
 
--- a/text/0034-extension-preconditions.md
+++ b/text/0034-extension-preconditions.md
@@ -1,0 +1,82 @@
+# Extension Preconditions
+- Start Date: 2026/09/04
+- RFC PR: https://github.com/openmrs/openmrs-rfc-frontend/pull/TBD
+
+## Decision, including impact on distributions
+
+An extension may declare, in its `routes.json` definition, the data it needs in order to have anything to render. The framework resolves that data before mounting and skips extensions whose requirement is not met, so an extension that would render nothing never becomes a parcel.
+
+Preconditions are developer-owned. Unlike display conditions, they are **not** configurable: implementers cannot add, alter, or remove them.
+
+Distributions require no changes. A definition without `preconditions` behaves exactly as it does today.
+
+## Definition
+
+Each extension the framework mounts is a single-spa parcel with its own root, to which React attaches its full set of delegated event listeners. Extensions that render conditionally on data they must fetch cannot make that decision until they have mounted, so this cost is paid whether or not anything is displayed.
+
+A **precondition** is a `{ requires, when }` object on an extension definition. `requires` names precondition providers to resolve before mounting. `when` is an optional expression refining the check, defaulting to *every required value is truthy*, and is evaluated against `{ session, ...slotState, ...resolvedValues }`, with resolved values taking precedence on a name collision.
+
+A **precondition provider** is a named function that resolves one value for one rendering of a slot. Providers are declared in a frontend module's `routes.json`, and that is the only way to register one: the framework exposes no public registration function. A distribution's providers are therefore always statically declared and discoverable by reading its modules' routes files, rather than appearing imperatively at arbitrary points during startup.
+
+The values a provider resolves are handed to the extension that required them, as a `preconditions` prop keyed by provider name. An extension gated on `activeVisit` reads it as `props.preconditions.activeVisit`; it does not fetch the visit a second time in order to display it. Data is usually loaded as a precondition precisely because the extension intends to render something from it, so gating on data the extension cannot then read would trade one wasted request for another.
+
+Preconditions are distinct from the **display conditions** of [RFC 27](0027-extensions.md). A display condition is implementer-owned policy — "hide this for male patients", "hide this for Nurses". A precondition is a developer's assertion that the extension has nothing to show without particular data.
+
+```json
+{
+  "extensions": [
+    {
+      "name": "visit-tag",
+      "component": "visitTag",
+      "preconditions": { "requires": ["activeVisit"], "when": "activeVisit != null" }
+    }
+  ],
+  "preconditionProviders": [
+    { "name": "activeVisit", "component": "activeVisitProvider" }
+  ]
+}
+```
+
+```ts
+type PreconditionResolver = (context: {
+  slotName: string;
+  state?: object;
+  session: Session | null;
+}) => unknown | Promise<unknown>;
+```
+
+Resolution proceeds as follows:
+
+- It runs *after* the existing synchronous filters — privileges, feature flags, online/offline, and display conditions — so only extensions that would otherwise mount are resolved.
+- The union of required names is resolved once per rendering of a slot and shared by every extension in that rendering which requires it.
+- An unregistered provider name, or a provider that throws or rejects, hides the extensions depending on it. This is also the correct outcome when a provider's frontend module is not installed in a distribution.
+- The resolved values are passed to the extension as its `preconditions` prop, and re-sent whenever the slot re-renders and they are resolved again. They are a snapshot taken to decide whether to mount, not a live subscription; see *Common practices* for how they stay current.
+
+`getAssignedExtensions` becomes asynchronous, retaining a synchronous path for extensions without preconditions. `useAssignedExtensions` keeps its array return type, yielding the synchronously-resolvable extensions immediately and appending the rest once resolved, so extensions without preconditions never wait.
+
+## Reason for decision
+
+- Measured on the reference application, one slot rendered once per row of a paginated list mounted 112 extensions across 15 rows, of which **82 (73%) rendered nothing at all** — roughly 10,700 delegated event listeners for no visible output.
+- Those same extensions issued about 30 REST requests per page change, one per row per extension, in order to discover that they had nothing to display.
+- The cost recurs on every re-render. In a paginated list, each page change repeats it in full.
+- Display conditions cannot express this: they are evaluated synchronously, so a condition cannot depend on fetched data.
+- Display conditions must not express this even if they could. Configuration *replaces* a definition's display expression outright, so an implementer setting a policy of their own would silently disable a correctness and performance guard they never knew was there.
+- Naming the required data lets the framework resolve it once per slot rendering rather than once per extension, and lets a provider batch across renderings — collapsing the per-row request pattern above to a single request per page.
+- Because the extension receives what its providers resolved, gating does not introduce a second fetch of the same data. Without this, an extension gated on the active visit would fetch it to decide and again to render it, and the mechanism would move requests around rather than remove them.
+
+## Alternatives
+
+- **Make display conditions asynchronous.** Conflates developer-owned data requirements with implementer-owned policy, and leaves the guard overrideable from configuration.
+- **Infer the required data from the expression** by extracting its unbound identifiers. Removes the `requires` list, but makes dependencies implicit and turns a misspelled provider name into a silently hidden extension rather than a reported error.
+- **Let each extension declare its own loader.** Most local, but the framework cannot deduplicate across extensions, so seven extensions needing the same visit still make seven requests.
+- **Pass everything through slot state.** The slot owner would have to know what its extensions need, coupling slots to their extensions and defeating the purpose of the extension system.
+- **Make empty renders cheaper**, for instance by sharing one root per slot. This would break the framework-agnostic parcel contract that allows extensions to be written in something other than React.
+- **Do nothing and remove the offending extensions by hand.** Does not generalize: the extensions are individually reasonable, and the next paginated slot reintroduces the problem.
+
+## Common practices (not enforced)
+
+- Prefer slot state to a provider when the data is already there. A precondition referring only to state and session resolves synchronously and costs nothing.
+- A provider serving a slot that renders once per row should coalesce its requests within a tick. The framework calls providers once per slot rendering; turning that into one request is the provider's responsibility.
+- Providers should fetch through SWR, as everything else does under [RFC 28](0028-data-updates.md). The `preconditions` prop then serves the first paint, and an extension that needs the data to stay current subscribes to the same key with its usual hook, which resolves from cache without another request and updates as normal. This keeps liveness in one mechanism rather than adding a second.
+- An extension that wants a provider's data but should always render can declare `requires` with `"when": "true"`. Declaring a data dependency and gating on it are separable, and the framework's deduplication and batching apply either way.
+- Preconditions answer "does this extension have anything to show?". Anything an implementer should be able to change belongs in a display condition instead.


text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+{
+  "extensions": [
+    {
+      "name": "visit-tag",
+      "component": "visitTag",
+      "preconditions": { "requires": ["activeVisit"], "when": "activeVisit != null" }
+    }
+  ],
+  "preconditionProviders": [
+    { "name": "activeVisit", "component": "activeVisitProvider" }
```
chibongho, line 35: Can we consider changing the terms `Providers` and `Component`? While this somewhat models React's Context providers, the provider function should just be a regular JS function. I prefer "fetcher", but open to other suggestions.

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Definition
+
+Each extension the framework mounts is a single-spa parcel with its own root, to which React attaches its full set of delegated event listeners. Extensions that render conditionally on data they must fetch cannot make that decision until they have mounted, so this cost is paid whether or not anything is displayed.
+
+A **precondition** is a `{ requires, when }` object on an extension definition. `requires` names precondition providers to resolve before mounting. `when` is an optional expression refining the check, defaulting to *every required value is truthy*, and is evaluated against `{ session, ...slotState, ...resolvedValues }`, with resolved values taking precedence on a name collision.
+
+A **precondition provider** is a named function that resolves one value for one rendering of a slot. Providers are declared in a frontend module's `routes.json`, and that is the only way to register one: the framework exposes no public registration function. A distribution's providers are therefore always statically declared and discoverable by reading its modules' routes files, rather than appearing imperatively at arbitrary points during startup.
```
chibongho, line 19: Are the following true?
- The precondition provider function takes in the extension's state as input.
- If multiple extensions require the same provider function, it will only run once, as the input to the function stays the same.

I would love this RFC to also solve the "1-request for all extensions instead of 1 request per extension" problem we have (in, for example, fetching the active visits for each patient in the patient search results). I think it's really close, but when I tried to picture it, I came with with a provider function like this that doesn't quite work (because the inputs are different for each specific patient):

```TS
// makes 1 request to fetch the activeVisit of all patients in the array, but only return the activeVisit of the specific patient
// the extension slot would have `{patientUuids, specificPatientUuid}` as its `state`
function getActiveVisits(patientUuids: Array<String>, specificPatientUuid: string) {
}
```

Maybe we can work around it by making the `specificPatientUuid` some sort of `key` of the extension slot, whose value doesn't get passed into the provider function?

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+Resolution proceeds as follows:
+
+- It runs *after* the existing synchronous filters — privileges, feature flags, online/offline, and display conditions — so only extensions that would otherwise mount are resolved.
+- The union of required names is resolved once per rendering of a slot and shared by every extension in that rendering which requires it.
+- An unregistered provider name, or a provider that throws or rejects, hides the extensions depending on it. This is also the correct outcome when a provider's frontend module is not installed in a distribution.
+- The resolved values are passed to the extension as its `preconditions` prop, and re-sent whenever the slot re-renders and they are resolved again. They are a snapshot taken to decide whether to mount, not a live subscription; see *Common practices* for how they stay current.
+
+`getAssignedExtensions` becomes asynchronous, retaining a synchronous path for extensions without preconditions. `useAssignedExtensions` keeps its array return type, yielding the synchronously-resolvable extensions immediately and appending the rest once resolved, so extensions without preconditions never wait.
```
jwnasambu, line 55: Kindly this appends gated extensions after the synchronous ones land so in a mixed slot the configured order can't be honored  a gated tag always renders after every ungated one. Render order follows the array order here so it's a silent reordering for slots where order is meaningful. Is "sync first, then async in resolution order" the intended semantics for mixed slots and do we want to note it in the document?

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+
+The values a provider resolves are handed to the extension that required them, as a `preconditions` prop keyed by provider name. An extension gated on `activeVisit` reads it as `props.preconditions.activeVisit`; it does not fetch the visit a second time in order to display it. Data is usually loaded as a precondition precisely because the extension intends to render something from it, so gating on data the extension cannot then read would trade one wasted request for another.
+
+Preconditions are distinct from the **display conditions** of [RFC 27](0027-extensions.md). A display condition is implementer-owned policy — "hide this for male patients", "hide this for Nurses". A precondition is a developer's assertion that the extension has nothing to show without particular data.
```
dkayiwa: RFC 27 does not define display conditions. Its Configuration section gives an extension slot exactly four keys, `add`, `remove`, `order` and `configure`, and the words condition, expression and privilege do not appear in it at all. No other RFC in `text/` mentions them either, so display conditions have never been through an RFC: they are a framework feature, defined by the reserved `Display conditions` config key in `esm-config` and documented in the o3 docs extension-system and configuration-system pages.

It matters here because the developer-owned versus implementer-owned split on this line is the hinge of the whole proposal, and two of the Reason bullets are claims about how display conditions behave. Could this point at where they are actually specified, or define them here? Someone who follows the link today lands on an extension system with no conditions in it.

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+```json
+{
+  "extensions": [
+    {
+      "name": "visit-tag",
+      "component": "visitTag",
+      "preconditions": { "requires": ["activeVisit"], "when": "activeVisit != null" }
+    }
+  ],
+  "preconditionProviders": [
```
dkayiwa: Are provider names meant to join the name-conflict problem RFC 32 left to a future RFC? They look global and cross-module here: `requires` names them bare, and line 52's not-installed case only makes sense if one module's extension can name another module's provider.

I am not asking this RFC to invent a rule the framework has nowhere else. `registerExtension`, `registerModal` and `registerWorkspace` are all silently last-write-wins today, and `routes.json` already carries eight bare-name arrays with no collision rule between them. I ask because a provider collision fails differently from those: the loser's extension is handed the wrong resolver, so it gets a wrong `props.preconditions.activeVisit` and is gated on data it never asked for, where a duplicate slot name at least warns and refuses to register. A sentence saying which way this one goes would settle it.

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+  state?: object;
+  session: Session | null;
+}) => unknown | Promise<unknown>;
+```
+
+Resolution proceeds as follows:
+
+- It runs *after* the existing synchronous filters — privileges, feature flags, online/offline, and display conditions — so only extensions that would otherwise mount are resolved.
+- The union of required names is resolved once per rendering of a slot and shared by every extension in that rendering which requires it.
+- An unregistered provider name, or a provider that throws or rejects, hides the extensions depending on it. This is also the correct outcome when a provider's frontend module is not installed in a distribution.
```
dkayiwa: This is the same outcome the Alternatives section uses to rule out inference on line 70, which faults it for turning "a misspelled provider name into a silently hidden extension rather than a reported error". As specified here, a misspelled name in `requires` is an unregistered provider name, so it hides the extension quietly too, and nothing in the document reports it anywhere. The explicit list is in fact what makes that error detectable, since a name in `requires` can only be a provider while an unbound identifier in an expression might just as well be a slot state key. The Definition then declines to use it.

The draft ticket has the missing half, "hides the dependent extensions and logs once". Could that land in this document, which is the part that gets merged? The routes loader already `console.error`s a definition missing a `name` or a `component`, and because both `requires` and the providers are static, the check can happen at registration rather than at render. Worth keeping the quiet hide for the genuinely-absent-module case in your second sentence, and reporting the rest.

text/0034-extension-preconditions.md
dkayiwa: This is a breaking change to a published API, and line 11 tells implementers otherwise. `getAssignedExtensions` sits in the curated public barrel (`esm-extensions/src/public.ts`, re-exported from `esm-framework/src/index.ts`), it is in the generated public surface in `docs/API.md`, the o3 docs teach it to app developers, and `.changeset/calm-slots-settle.md` currently tells module authors to call it rather than read the store. Two apps in the org do: `openmrs-esm-dispensing-app` at `src/prescriptions/prescription-expanded.component.tsx:19`, which gates a tab on `.length > 0`, and `openmrs-esm-patient-chart` at `packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx:96`, which builds the chart's dashboard list inside a `flatMap`. The second is the patient chart case you raise in the PR description, and an org-wide search turns up no others.

What I would want stated is that the failure mode is not a build error. `pinFrameworkToAppShell` registers a Module Federation `resolveShare` plugin that forces every module's `@openmrs/esm-framework` to the shell's copy whatever version it was built against, so a distribution that assembles today's published dispensing-app and patient-chart onto a shell carrying this change runs the new async function inside old compiled code: dispensing's Conditions and diagnoses tab silently stops appearing, since `undefined > 0` is false, and chart review throws on `.map`. That lands on a distribution which changed nothing, and line 11 is the sentence promising it cannot.

RFC 32 and RFC 33 both disclose this sort of thing in the same section, 33 with "Modules adopt this with a tooling bump" and 32 with its two transition-window sentences. Could line 11 grow that companion sentence, naming the two apps and the release this rides? Keeping the export synchronous is not the way out, for what it is worth: a sync view that skips the precondition filter is the hazard `calm-slots-settle` renamed `candidateExtensions` to prevent, and dispensing would then render an empty tab, which is the waste this RFC exists to remove.

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+- Naming the required data lets the framework resolve it once per slot rendering rather than once per extension, and lets a provider batch across renderings — collapsing the per-row request pattern above to a single request per page.
+- Because the extension receives what its providers resolved, gating does not introduce a second fetch of the same data. Without this, an extension gated on the active visit would fetch it to decide and again to render it, and the mechanism would move requests around rather than remove them.
+
+## Alternatives
+
+- **Make display conditions asynchronous.** Conflates developer-owned data requirements with implementer-owned policy, and leaves the guard overrideable from configuration.
+- **Infer the required data from the expression** by extracting its unbound identifiers. Removes the `requires` list, but makes dependencies implicit and turns a misspelled provider name into a silently hidden extension rather than a reported error.
+- **Let each extension declare its own loader.** Most local, but the framework cannot deduplicate across extensions, so seven extensions needing the same visit still make seven requests.
```
dkayiwa: Not under SWR, and RFC 28 makes SWR how everything fetches. In swr 2 (2.2.5 is what the repo resolves today), `shouldStartNewRequest = !FETCH[key] || !opts.dedupe`, so a call for a key already in flight joins the one in progress rather than issuing its own, and `preload` does the same through its own `PRELOAD[key]` check. Seven extensions loading the same visit under the same key make one request, not seven.

The difference that does hold is the one you put in the PR description, that resolving in the framework need not rely on SWR to deduplicate, and that a provider is resolved once per slot rendering however it fetches. Could the bullet say that instead? As written it undersells the alternative by a factor of seven, and this is the section people quote back later.

text/0034-extension-preconditions.md
```diff
@@ -0,0 +1,82 @@
...
+- **Pass everything through slot state.** The slot owner would have to know what its extensions need, coupling slots to their extensions and defeating the purpose of the extension system.
+- **Make empty renders cheaper**, for instance by sharing one root per slot. This would break the framework-agnostic parcel contract that allows extensions to be written in something other than React.
+- **Do nothing and remove the offending extensions by hand.** Does not generalize: the extensions are individually reasonable, and the next paginated slot reintroduces the problem.
+
+## Common practices (not enforced)
+
+- Prefer slot state to a provider when the data is already there. A precondition referring only to state and session resolves synchronously and costs nothing.
```
dkayiwa: Is `requires` optional? Line 17 defines a precondition as `{ requires, when }` with only `when` marked optional, but a precondition "referring only to state and session" has no provider in it, while the `"when": "true"` bullet below needs one.

If a bare `when` is allowed, then this mechanism is also a synchronous display condition that implementers cannot override, and that is a bigger change than the Decision section lets on: today an implementer can replace any `displayExpression` through the `Display conditions` config key, and after this a developer can put the same gate somewhere they cannot reach. The closing bullet of this section reads as though that is deliberate, in which case it belongs up in the Decision, since it moves the implementer/developer boundary rather than adding a data mechanism. If it is not deliberate, saying `requires` must be non-empty closes it.

text/0034-extension-preconditions.md
dkayiwa: The part that makes this land is the batching, and I do not think the document commits to it. Line 64 offers "collapsing the per-row request pattern above to a single request per page" as a reason for the decision, but the Common practices bullet puts the coalescing on the provider author and marks it not enforced, and the resolver signature carries only `{ slotName, state, session }` with no batch context. Your example does work without a `key` concept, I think: the resolver is called once per rendering with that rendering's own state, so a provider can push `state.patientUuid` into a batch, flush on a microtask and hand each caller back its own slice. But that is dataloader plumbing each provider author would rewrite.

The draft ticket says it out loud: "Also provide an optional batching helper so a provider can collapse per-row resolutions within a tick into one request. This is what removes the N+1 fetching; gating alone does not." That helper is what you are asking for, and it is the load-bearing piece of the 30-requests-per-page-change figure in the Reason section, so it seems worth having in the RFC rather than only in the ticket. Failing that, line 64 should say the collapse depends on each provider implementing it.

text/0034-extension-preconditions.md
ibacher: Happy to discuss naming! Definitely none of this is set in stone.

On `component`, this is just mirroring how we already refer to named exports from an app's entrypoint. Granted, it makes more sense in the context where those usually are literally React components. But, we could call it `function` or `export` or something if that's clearer?

On `provider`, I was hedging a bit. Technically, whatever we call them, these are just named functions that somehow determine a value to set for a precondition expression. The idea is to use them for data fetching. What if we called them "Loaders" or something like that?

text/0034-extension-preconditions.md
ibacher: > The precondition provider function takes in the extension's state as input.

Yes.

> If multiple extensions require the same provider function, it will only run once, as the input to the function stays the same.

Strictly speaking, not part of this. The main limitation here is just that the REST API really doesn't support (yet!) cross-patient searching like this. However, Claude suggested something like this utility:

```ts
function createBatcher(fetchMany, { windowMs = 0, maxBatch = 50, keyFn = String } = {}) {
  const pending = new Map();   // key -> { resolve, reject, input }
  const inFlight = new Map();  // key -> promise
  let scheduled = false;

  function flush() {
    scheduled = false;
    const entries = [...pending];
    pending.clear();

    for (let i = 0; i < entries.length; i += maxBatch) {
      const chunk = entries.slice(i, i + maxBatch);
      Promise.resolve(fetchMany(chunk.map(([, e]) => e.input))).then(
        (results) => chunk.forEach(([, e], j) => e.resolve(results[j])),
        (err) => chunk.forEach(([, e]) => e.reject(err))
      );
    }
  }

  return function load(input) {
    const key = keyFn(input);
    if (inFlight.has(key)) return inFlight.get(key);

    const promise = new Promise((resolve, reject) => {
      pending.set(key, { resolve, reject, input });
    }).finally(() => inFlight.delete(key));

    inFlight.set(key, promise);

    if (!scheduled) {
      scheduled = true;
      windowMs === 0 ? queueMicrotask(flush) : setTimeout(flush, windowMs);
    }
    return promise;
  };
}
```

Then it becomes something like:

```ts
const getActiveVisit = createBatcher(async (uuids) => {
  // this search needs to be built
  const visits = (await openmrsFetch(`/visits?patients=${ids.join(",")}&activeOnly=true&v=custom:(uuid,patient:(uuid),...)`)).data.data;
  return uuid.map((uuid) => visits.find((v) => v.patient?.uuid === uuid) ?? null);
}, { windowMs = 10 });
```

The above needs some work for resiliency and so forth, but something like that. The idea here is we have a (short!) window over which calls are queued and once that window expires, we run the query and return the results and then handle the next batch, etc.

text/0034-extension-preconditions.md
ibacher: That's a good point! I think this is reviseable saying something like: instead of `Extension` objects we return an array of `Promise<Extension>` objects. For synchronously-defined extensions, these are immediately mountable whereas asynchronously mounted ones are only available after checking the data preconditions, but that way the array retains the correct order.

ibacher: This does force the whole extension setup to be asynchronous, but it kind of already is.

NethmiRodrigo (approved):

